PR #100, opened 2015-01-12
Added setup for a windows based build

dpwspoon:
Not sure how whether we want a second build badge in the README, but we do want a Windows machine because we had a bind error in the past


dpwspoon:
Merged this branch with other bug fix:  Appveyor was trying to build on new branch and failing because appveyor.yml was missing.

PR #115, opened 2015-01-28
Moved HttpRobotBehaviorIT to itest module

dpwspoon:


jfallows:
Looks like RobotIT is still non-deterministic.
Suggest moving RobotIT to itest module as well.

dpwspoon:
Agreed, refactoring these tests and RobotIT especially in conjunction with k3po/k3po#117 should help to clear up the travis-ci builds.  I'm going to close the pull request for now and see if we can work towards that end.

jfallows:
That's even better, thanks @dpwspoon !
Suggest removing the `itest` sub-module completely as part of k3po/k3po#117 too.

Issue #101, opened 2015-01-12
HttpServerBootstrapTest#shouldAcceptEchoThenClose occasionally failing on TravisCI


dpwspoon:
```
Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.21 sec <<< FAILURE!
shouldAcceptEchoThenClose(org.kaazing.k3po.driver.netty.bootstrap.http.HttpServerBootstrapTest)  Time elapsed: 0.992 sec  <<< FAILURE!
org.junit.experimental.theories.internal.ParameterizedAssertionError: shouldAcceptEchoThenClose("CHUNKED" <from CONTENT_STRATEGIES[1]>)
	at org.junit.experimental.theories.Theories$TheoryAnchor.reportParameterizedError(Theories.java:288)
	at org.junit.experimental.theories.Theories$TheoryAnchor$1$1.evaluate(Theories.java:237)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithCompleteAssignment(Theories.java:218)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:204)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithIncompleteAssignment(Theories.java:212)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:202)
	at org.junit.experimental.theories.Theories$TheoryAnchor.evaluate(Theories.java:187)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:252)
[...]

PR #125, opened 2015-02-11
Fixed bug where HTTP requests and responses could not be only half specified

dpwspoon:


dpwspoon:
The test failures on appveyor are a replication of k3po/k3po#101, so it should be safe to merge

PR #133, opened 2015-02-18
Removed itests module as all the tests have been moved to driver And Fix #134 (Regex variable capture type)

### LINKED ISSUE #134: A variable captured via a regex capture should be usable as an int.
Thus we could do something like

```
accept http://localhost:8080/
accepted
connected

read method "POST"
read header "Content-length" /(?<length>.*)/
read [0..${length}]
read closed

write status "200" "OK"
write header content-length
write close

```

dpwspoon:
This somehow got omitted from prior work


dpwspoon:
Added fix for #135

PR #140, opened 2015-03-04
K3PO Launcher

pkhanal:
- Added a module k3po.launcher that generates the exectuable jar


launcher/src/main/java/org/kaazing/k3po/launcher/Launcher.java
```diff
@@ -0,0 +1,88 @@
...
+        } catch (ParseException ex) {
+            HelpFormatter helpFormatter = new HelpFormatter();
+            helpFormatter.printHelp(Launcher.class.getSimpleName(), options, true);
+        }
+
+    }
+
+    private static Options createOptions() {
+        Options options = new Options();
+        Option scriptPath = new Option(null, "scriptPath", true, "path to directory/jar for script(s) lookup");
```
dpwspoon: If this is for jars as well, than I think it may be the case that we need multiple paths...?

launcher/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+                <artifactId>maven-assembly-plugin</artifactId>
+            </plugin>
+        </plugins>
+        
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-assembly-plugin</artifactId>
+                    <version>2.4</version>
```
dpwspoon: I could go either way, but just want to ask:  Do we want to create a consistent name (i.e. not attach the version to the name of the jar when it is generated)?

dpwspoon:
Looks good!  Let me know what you think for the questions.

dpwspoon:
Also this could potentially go in the driver module.  Maven would then generate 2 artifacts which may or may not be an anti-pattern

launcher/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+                    <configuration>
+                        <descriptorRefs>
+                            <descriptorRef>jar-with-dependencies</descriptorRef>
+                        </descriptorRefs>
+                        <appendAssemblyId>false</appendAssemblyId>
+                        <archive>
+                            <manifest>
+                                <mainClass>org.kaazing.k3po.launcher.Launcher</mainClass>
```
dpwspoon: It would be nice for the command line to have an option to list the version.  You can add [it to the manifest](http://maven.apache.org/shared-archives/maven-archiver-2.5/examples/manifest.html) and then access it from java

launcher/src/main/java/org/kaazing/k3po/launcher/Launcher.java
jfallows: We could have `--scriptPath` behave similarly to `--classpath` for `java` which separates entries via semicolons, where each entry can be a directory or a JAR file.

`java --classpath .;/path/to/libs/library.jar ...`

For consistency, perhaps it should be `--scriptpath` instead of `--scriptPath`.

pkhanal: Done!

pkhanal:
@dpwspoon, Thanks!

Issue #145, opened 2015-03-16
Multiple spaces between hex characters cause script parse errors


cmebarrow:
Syntax like the following is rejected:

write [0x81  0xfe  0x00]

The error message is:
> [ERROR] Caught exception trying to parse script. Sending error to client. Due to org.kaazing.k3po.lang.parser.ScriptParseException:  
> org.kaazing.k3po.lang.parser.ScriptParseException: Syntax error while parsing:
line 42:6 token recognition error at: '[0x81  '

cmebarrow:
This used to work in the previous version (robot).

dpwspoon:
Should be an easy fix to Robot.g4

```
BytesLiteral
    : '[' (' ')? (ByteLiteral (' ')?)+ ']'
    ;
```

will become

```
BytesLiteral
    : '[' (' ')? (ByteLiteral (' ')*)+ ']'
    ;
```

cmebarrow:
I'll take a shot at it

PR #146, opened 2015-03-16
#145: minor change to grammar to allow multiple spaces between hex chara...

cmebarrow:
...cters in byte literals.


dpwspoon:
closes #145

Issue #191, opened 2015-06-12
Specification.ws: SocketException with Broken Pipe as connection is closed prematurely


sanjay-saxena:
When `ClosingHandshakeIT.shouldCompleteCloseHandshakeWhenServerSendCloseFrameWithCode1000AndReasonUsingMessageReader()` test is executed using `server.send.close.frame.with.code.1000.and.reason/handshake.response.and.frames` script, the following exception is seen sporadically:

```
shouldCompleteCloseHandshakeWhenServerSendCloseFrameWithCode1000AndReasonUsingMessageReader(org.kaazing.netx.ws.specification.ClosingHandshakeIT)  Time elapsed: 0.029 sec  <<< ERROR!
java.net.SocketException: Broken pipe
	at java.net.SocketOutputStream.socketWrite0(Native Method)
	at java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:113)
	at java.net.SocketOutputStream.write(SocketOutputStream.java:138)
	at org.kaazing.netx.http.internal.HttpURLConnectionHandler$TcpOutputStream.write(HttpURLConnectionHandler.java:484)
	at org.kaazing.netx.ws.internal.io.OutgoingSentinelExtension.encodeMaskAndPayload(OutgoingSentinelExtension.java:300)
	at org.kaazing.netx.ws.internal.io.OutgoingSentinelExtension.validateAndEncodeCloseFrame(OutgoingSentinelExtension.java:269)
	at org.kaazing.netx.ws.internal.io.OutgoingSentinelExtension.access$100(OutgoingSentinelExtension.java:50)
	at org.kaazing.netx.ws.internal.io.OutgoingSentinelExtension$3.accept(OutgoingSentinelExtension.java:82)
	at org.kaazing.netx.ws.internal.ext.WebSocketContext.onCloseSent(WebSocketContext.java:125)
	at org.kaazing.netx.ws.internal.WebSocketOutputStateMachine.processFrame(WebSocketOutputStateMachine.java:85)
	at org.kaazing.netx.ws.internal.WsURLConnectionImpl.processOutgoingFrame(WsURLConnectionImpl.java:730)
	at org.kaazing.netx.ws.internal.io.WsOutputStream.writeClose(WsOutputStream.java:189)
	at org.kaazing.netx.ws.internal.WsURLConnectionImpl.sendClose(WsURLConnectionImpl.java:872)
[...]

PR #192, opened 2015-06-12
Issue #191: Added the missing line for the server to read the status …

sanjay-saxena:
…code and the reason so that the server does not close the connection while the client is still writing.


jfallows:
Looks good, @sanjay-saxena.

PR #193, opened 2015-06-12
Updated pom.xml to pickup the latest community also added animal-snif…

sanjay-saxena:
…fer property for java17 byte code.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
     </parent>
 
     <artifactId>k3po.parent</artifactId>
-    <version>2.0.1</version>
+    <version>develop-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -6,11 +6,11 @@
     <parent>
         <groupId>org.kaazing</groupId>
         <artifactId>community</artifactId>
-        <version>2.11</version>
+        <version>2.13</version>
     </parent>
 
     <artifactId>k3po.parent</artifactId>
-    <version>develop-SNAPSHOT</version>
+    <version>2.0.1</version>
```
dpwspoon: This is not correct for develop branch

Issue #195, opened 2015-06-16
Specification.ws: Add new tests and scripts to FragmentationIT for client and server


sanjay-saxena:
to exchange fragmented frames. The current tests have one side sending fragmented frames and the other side echoing a complete message. We need to add tests(for both binary and text) such that both sides dealing with fragmented frames.

sanjay-saxena:
We will reuse the existing test scripts as they should be enough to deal with both -- message in a single WebSocket frame and message split across multiple WebSocket frames.

PR #196, opened 2015-06-16
Issue #195: Added new tests to FragmentationIT so that both sides can…

sanjay-saxena:
… send and receive fragmentated binary and text frames.


jfallows:
Hey @sanjay-saxena, the tests for fragmented data payloads (binary and text) are all test scenarios of the form `client.echo.*` and `server.echo.*`, whereas other non-data frames are verified in test scenarios of the form `client.send.*` and `server.send.*`.

The new files added for test scenarios `client.send.binary.payload.length.125.fragmented` and `server.send.binary.payload.length.125.fragmented` seem like they might almost be covered by test scenarios `client.echo.binary.payload.125.fragmented` and `server.echo.binary.payload.125.fragmented`.

Please take a look and see if the diffs between the existing scripts and the new scripts are fundamental to the new test scenarios, or if there is enough overlap to warrant consolidating into a single test scenario instead.

sanjay-saxena:
Agreed @jfallows. It looks like we can reuse the existing scripts. I will close this PR.

Issue #187, opened 2015-05-11
MessageFormatIT#inboundShouldAcceptHeaders sporadically failing


dpwspoon:
from [travis build](https://s3.amazonaws.com/archive.travis-ci.org/jobs/62115509/log.txt)

PR #206, opened 2015-07-20
Updated the wseb version to 1.0 to enable testing against current implementation

pkhanal:
Also fixed the strictness in the scripts those were expecting 'read closed' immediately after response with status code 400

### DIFF
--- a/specification/wse/src/main/scripts/org/kaazing/specification/wse/browsers/client.reconnect.downstream/request.rpt
+++ b/specification/wse/src/main/scripts/org/kaazing/specification/wse/browsers/client.reconnect.downstream/request.rpt
@@ -7,14 +7,14 @@ write method "POST"
 write version "HTTP/1.1"
 write header host
 write header content-length
-write header "X-WebSocket-Version" "wseb-1.1"
+write header "X-WebSocket-Version" "wseb-1.0"
 write header "X-Sequence-No" ${wse:asString(sequence)}
 write close
 
 read status "201" /.+/
 read version "HTTP/1.1"
-read header "X-WebSocket-Version" "wseb-1.1"
-read header "Content-Type" "text/plain;charset=utf-8"
+read header "X-WebSocket-Version" "wseb-1.0"
+read header "Content-Type" /text\/plain;charset=(?i)utf-8/
 
 read /(?<upstream>http:\/\/localhost:8080\/path\/.+)/ "\n"
 read /(?<downstream>http:\/\/localhost:8080\/path\/.+)/ "\n"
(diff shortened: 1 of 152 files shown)


specification/wse/src/main/scripts/org/kaazing/specification/wse/browsers/client.reconnect.downstream/request.rpt
```diff
@@ -36,22 +36,20 @@ write close
 read status "200" /.+/
 read header "Content-Type" "application/octet-stream"
 read header "Connection" "close"
+read notify DOWNSTREAM_OPENED
 
-read [0x80 0x82 0x00]
-read [0..256]
-read notify DATA_RECEIVED
```
cmebarrow: Why did you remove this? Just because it's unnecessary? (which is a fine reason, just wanted to know if there was any other)

specification/wse/src/main/scripts/org/kaazing/specification/wse/browsers/client.reconnect.downstream/request.rpt
```diff
@@ -7,14 +7,14 @@ write method "POST"
...
+write header "X-WebSocket-Version" "wseb-1.0"
 write header "X-Sequence-No" ${wse:asString(sequence)}
 write close
 
 read status "201" /.+/
 read version "HTTP/1.1"
-read header "X-WebSocket-Version" "wseb-1.1"
-read header "Content-Type" "text/plain;charset=utf-8"
+read header "X-WebSocket-Version" "wseb-1.0"
+read header "Content-Type" /text\/plain;charset=(?i)utf-8/
```
cmebarrow: Why do we need the "?i"? Does the server add a space?

cmebarrow:
@pkhanal LGTM, just a couple of minor questions (see my comments). Do you think we should add a negative test for sequencing (i.e. out of order request)? Spec says that should result in 400 bad request.

specification/wse/src/main/scripts/org/kaazing/specification/wse/browsers/client.reconnect.downstream/request.rpt
pkhanal: I wanted to make the test as much tight as possible without losing the purpose of the test.

specification/wse/src/main/scripts/org/kaazing/specification/wse/browsers/client.reconnect.downstream/request.rpt
pkhanal: It's just to make the charset value case insensitive. I have seen cases where server would return "text/plain;charset=UTF-8". Semantically it is equivalent to "text/plain;charset=utf-8". It may not be robust for all combinations but we can fix that as we come across those scenarios.

pkhanal:
@cmebarrow, we already have negative tests for sequencing. They are available for [upstream requests](https://github.com/k3po/k3po/tree/develop/specification/wse/src/main/scripts/org/kaazing/specification/wse/upstream/request.out.of.order) and [downstream](https://github.com/k3po/k3po/tree/develop/specification/wse/src/main/scripts/org/kaazing/specification/wse/downstream/request.out.of.order) requests.

cmebarrow:
@pkhanal OK, thanks for your answers. Could you look into the build failure? Test MessageFormatIT.inboundShouldAcceptHeaders failed on openJdk. Could be a sporadic failure. If so please file a bug and rebuild.

pkhanal:
@cmebarrow, there is already an [issue](https://github.com/k3po/k3po/issues/187) filed for the failure you pointed.

cmebarrow:
OK, merging

PR #209, opened 2015-07-31
Add fin bit to fragmentation spec test

chadpowers:
@jitsni Add so that gateway will correctly echo back once receives end of message.


dpwspoon:
LGTM, thanks @chadpowers

PR #213, opened 2015-08-28
File transport

jitsni:
NOT ready for merge or review.

@jfallows @cmebarrow Seeking feedback on grammar changes. So review the following files:
1. Robot.g4
2.  file.connect.write.rpt


lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
```diff
@@ -41,7 +41,11 @@ acceptableNode
     ;
 
 connectNode
-    : k=ConnectKeyword (AwaitKeyword barrier=Name ConnectKeyword)? connectURI=location (OptionKeyword TransportKeyword value=location)? streamableNode+
+    : k=ConnectKeyword (AwaitKeyword barrier=Name ConnectKeyword)? connectURI=location
+                       (OptionKeyword TransportKeyword value=location)?
+                       (OptionKeyword SizeKeyword DecimalLiteral)?
+                       (OptionKeyword ModeKeyword ModeValue)?
+        streamableNode+
     ;
```
jitsni: 1. There is order in the options (may be better not to have it but adds additional nodes in the grammar).
2. If the mode is "rw", the file is created if there isn't one.
3. Size is used while creating the mapped byte buffer. It is optional when the file exists, but required when the file is created.

k3po-maven-plugin/src/main/java/org/kaazing/k3po/maven/plugin/internal/StartMojo.java
```diff
@@ -52,7 +52,7 @@
     /**
      * @parameter name="control" default-value="tcp://localhost:11642"
      */
-    private URI controlURI;
+    private URI control;
```
cmebarrow: Why this change? It loses information

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
```diff
@@ -57,10 +61,10 @@ optionNode
...
 
 writeOptionNode: 
-    k=WriteKeyword OptionKeyword name=MaskKeyword value=writeValue;
+    k=WriteKeyword OptionKeyword (name=MaskKeyword value=writeValue | name=OffsetKeyword DecimalLiteral);
```
cmebarrow: I think you need to put value=DecimalLiteral

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
cmebarrow: If mode is "r" that suggests you can't write, but I presume that's not your intention? If the intention is simply to specify whether or not the file is created, maybe just have "connection option create"?

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
jitsni: I would be putting offset=DecimalLiteral so that it won't clash with "writeValue". But that's only for processing it doesn't affect the syntax.

k3po-maven-plugin/src/main/java/org/kaazing/k3po/maven/plugin/internal/StartMojo.java
jitsni: I had to move to maven-plugin-plugin to 3.4 (because of jdk8) and that seems to have broken this.

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
jitsni: Yes, that was the intention. "r" is for read only mode. We open the file and this parameter corresponds to mode parameter in RandomAccessFile(File, String mode). I think this would also protect against accidentally modifying the file. Otherwise, we would be opening the file in read-write mode always even if the script just wants to do read operations.

Regarding whether we want to have an explicit option for creation, I thought "rw" mode can be used just like RandomAccessFile.

Other thing is whether we want the script to be more expressive instead of "r", "rw". We can have something like "read-only", "read-write" etc. But I think mode is very common.

cmebarrow: if "r" means writes are forbidden then I am fine with it.

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
jitsni: I think it is better to split each option into its own node:

```
writeOptionNode
    : writeOptionMaskNode
    | writeOptionOffsetNode
    ;
    
writeOptionMaskNode
    : k=WriteKeyword OptionKeyword name=MaskKeyword value=writeValue
    ;
    
writeOptionOffsetNode
    : k=WriteKeyword OptionKeyword name=OffsetKeyword value=DecimalLiteral
    ;
```

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
```diff
@@ -41,7 +41,11 @@ acceptableNode
     ;
 
 connectNode
-    : k=ConnectKeyword (AwaitKeyword barrier=Name ConnectKeyword)? connectURI=location (OptionKeyword TransportKeyword value=location)? streamableNode+
+    : k=ConnectKeyword (AwaitKeyword barrier=Name ConnectKeyword)? connectURI=location
+                       (OptionKeyword TransportKeyword value=location)?
```
dpwspoon: You could put each one of these in a node, right? i.e. locationOption maps to (OptionKeyword TransportKeyword value=location), or something like that

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
dpwspoon: +1 on separate nodes

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
dpwspoon: In general LGTM, but I'm not super familiar with random access file API.  One concern is that we are adding grammar nodes that might not make sense.  e.g. `connect tcp://localhost option mode rw` doesn't make sense for tcp right?  

This is sort of an outstanding issue with the grammar anyways (you can have http keywords in tcp and it still compiles, though you get other exceptions when k3po tries to setup the pipeline).  So bonus points if you can have the option only available in the case it is a file uri.

jitsni:
Ready for one more round of review.

lang/src/main/antlr4/org/kaazing/k3po/lang/parser/v2/Robot.g4
jitsni: We also don't want the options to be repeated. If we can write the grammar, that would be good.

dpwspoon:
@jfallows please review

jfallows:
Recommend keeping the `AstReadOptionNode` and `AstWriteOptionNode` generic, following the same approach as `AstReadConfigNode` and `AstWriteConfigNode`.

That will eliminate the need to modify so many interfaces, and let us more easily move to an extensible grammar for different protocol schemes in future.

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannel.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    protected boolean setClosed() {
+        return super.setClosed();
+    }
+
+    @Override
+    protected void setLocalAddress(ChannelAddress localAddress) {
+        super.setLocalAddress(localAddress);
+    }
```
jfallows: Should these overridden methods be `protected` scope or `package-private` scope?

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannelFactory.java
```diff
@@ -0,0 +1,37 @@
...
+package org.kaazing.k3po.driver.internal.netty.bootstrap.file;
+
+import org.jboss.netty.channel.ChannelFactory;
+import org.jboss.netty.channel.ChannelPipeline;
+
+public class FileChannelFactory implements ChannelFactory {
+
+    @Override
+    public FileChannel newChannel(ChannelPipeline pipeline) {
+        return new FileChannel(this, pipeline, new FileChannelSink());
```
jfallows: Perhaps the `FileChannelSink` should be one per factory, not one per channel?

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannelSink.java
```diff
@@ -0,0 +1,181 @@
...
+
+public class FileChannelSink extends AbstractChannelSink {
+
+    private static final InternalLogger LOGGER = InternalLoggerFactory.getInstance(FileChannelSink.class);
+
+    // Note that read position and write position are independent in the script. We are keeping
+    // two buffers(view on the same MappedByteBuffer) to achieve this and also don't have to
+    // manage positions explicitly
+    private ChannelBuffer readBuffer;
+    private ChannelBuffer writeBuffer;
```
jfallows: Shouldn't this buffer state be per-channel, not per channel sink?

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannelSink.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    protected void writeRequested(ChannelPipeline pipeline, MessageEvent evt) throws Exception {
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("writeRequested pipeline = " + pipeline + " evt = " + evt);
+        }
+        ChannelBuffer channelBuffer = (ChannelBuffer) evt.getMessage();
+        writeBuffer.writeBytes(channelBuffer);
+        ChannelFuture writeFuture = evt.getFuture();
+        writeFuture.setSuccess();
```
jfallows: What happens if there is insufficient space at the end of the file to accommodate the write request?

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/channel/ChannelAddress.java
```diff
@@ -50,6 +52,7 @@ public int compare(ChannelAddress o1, ChannelAddress o2) {
     private final boolean ephemeral;
 
     private final ChannelAddress transport;
+    private final Map<String, Object> options;
```
jfallows: Recommend keeping generic options out of the general purpose ChannelAddress API, and instead add specific options strongly typed to FileChannelAddress since they are only consumed by the file transport.

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannelSink.java
jitsni: IndexOutOfBoundsException is thrown since it is written to ChannelBuffer

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannel.java
jitsni: Since these are overridden methods, we cannot weaken the scope

jitsni:
Ready for one more round of review

jfallows:
Looks good @jitsni !

PR #216, opened 2015-09-10
Fixed race condition in read vs write parameters

dpwspoon:


dpwspoon:
@jitsni Can you please review?

jitsni:
looks good to me

PR #217, opened 2015-09-10
Added specification tests for SOCKS5+ protocol extension

alingherasim:
Added the socks5plus project and the specification tests which cover the CONNECT and BIND requests of the protocol.


specification/socks5plus/SPEC.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Abstract
+
+This document specifies an extended form of the standard SOCKS 5 protocol used for Enterprise Shield (a.k.a. reverse connectivity). This is sometimes referred to as SOCKS 5+.
+
+## Requirements
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119][].
+
+## The Protocol
+Kaazing SOCKS 5+ protocol conforms to [SOCKS 5] except that an extra ATYP (address type) is supported, as follows:
```
dpwspoon: Please add link to socks 5 for the brackets part

specification/socks5plus/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+    <reporting>
+        <plugins>
```
dpwspoon: There is no source code so I don't see the reason to include this here.  Ideally, it would be configured up in the parent if you really wanted to add it

specification/socks5plus/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    <description>SOCKS5 Plus Protocol specification</description>
+    <packaging>jar</packaging>
+
+    <scm>
+        <url>https://github.com/k3po/specification.socks5plus.git</url>
+        <connection>scm:git:https://github.com/k3po/specification.socks5plus.git</connection>
+    </scm>
+
+    <properties>
+        <maven.compiler.source>1.7</maven.compiler.source>
```
dpwspoon: This should be inherited, correct?

specification/socks5plus/src/main/java/org/kaazing/specification/socks5plus/Dummy.java
```diff
@@ -0,0 +1,23 @@
+package org.kaazing.specification.socks5plus;
```
dpwspoon: Shouldn't need Dummy class with package-info

specification/socks5plus/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+            <artifactId>k3po.junit</artifactId>
+           <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
```
dpwspoon: Shouldn't need src/main/resources as everything is in scripts

specification/socks5plus/src/test/java/org/kaazing/specification/socks5plus/request/bind/Socks5BindHandshakeIT.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.rules.TestRule;
+import org.junit.rules.Timeout;
+import org.kaazing.k3po.junit.annotation.Specification;
+import org.kaazing.k3po.junit.rules.K3poRule;
+
+/**
+ * SOCKS5 - RFC 1928, section 4 "Requests" and section 6 "Replies"
+ */
+public class Socks5BindHandshakeIT {
+    private final K3poRule k3po = new K3poRule().setScriptRoot("org/kaazing/specification/socks5plus/request");
```
dpwspoon: This seems like an odd path with the "request", since each script has a request and response.  Probably just want org/kaazing/specification/socks5plus or org/kaazing/specification/socks5plus/bind

specification/socks5plus/src/test/java/org/kaazing/specification/socks5plus/request/bind/Socks5BindHandshakeIT.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private final TestRule timeout = new DisableOnDebug(new Timeout(5, SECONDS));
+
+    @Rule
+    public final TestRule chain = outerRule(k3po).around(timeout);
+
+    @Test
+    @Specification({
+        "bind/connection.request.bind.with.uri.valid.port.succeeded.request",
```
dpwspoon: The folder prior to this "bind" should either be in the setScriptRoot (Thereby grouping all the tests that are in this class).  Additionally, you MAY want another folder here so the scripts are (`connection.request.bind.with.uri.valid.port.succeeded/request` and `connection.request.bind.with.uri.valid.port.succeeded/response`).  That's the template generally used by the other specification projects as it groups the scripts together

specification/socks5plus/src/test/java/org/kaazing/specification/socks5plus/request/connect/Socks5ConnectHandshakeIT.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Rule
+    public final TestRule chain = outerRule(k3po).around(timeout);
+
+    @Test
+    @Specification({
+        "connect/connection.request.connect.with.uri.valid.port.succeeded.request",
+        "connect/connection.request.connect.with.uri.valid.port.succeeded.response"
+        })
+    public void shouldConnectWithUriAndValidPortSucceeded() throws Exception {
```
dpwspoon: We should have a test for the special case port (all zeros if it doesn't exist) as well as a random normal port

specification/socks5plus/SPEC.md
```diff
@@ -0,0 +1,28 @@
+# Kaazing SOCKS 5+ protocol
+
+## Abstract
+
+This document specifies an extended form of the standard SOCKS 5 protocol used for Enterprise Shield (a.k.a. reverse connectivity). This is sometimes referred to as SOCKS 5+.
+
+## Requirements
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119][].
```
dpwspoon: please add link

specification/socks5plus/SPEC.md
```diff
@@ -0,0 +1,28 @@
...
+
+## The Protocol
+Kaazing SOCKS 5+ protocol conforms to [SOCKS 5] except that an extra ATYP (address type) is supported, as follows:
+
+URI address: X'FE'
```
dpwspoon: This is OK, but it would be nice if it was a little more close to the style of the RFC specification.  For example it wasn't clear to me until reading the scripts that the AYTP we are reserving is "0xFE" (The X was confusing)

specification/socks5plus/src/main/scripts/org/kaazing/specification/socks5plus/request/bind/connection.request.bind.with.invalid.uri.request.rpt
```diff
@@ -0,0 +1,69 @@
...
+
+# RSV field
+write [0x00]
+
+# ATYP with URI
+write [0xFE]
+
+# DST.ADDR with invalid URI
+# URI length
+write [0x00 0x00 0x00 0x09]
```
jitsni: I see that gateway is sending this. Is this correct ? For variable length address, the first byte gives the length, but here you are using 4 bytes.

specification/socks5plus/src/main/scripts/org/kaazing/specification/socks5plus/request/bind/connection.request.bind.with.invalid.uri.request.rpt
```diff
@@ -0,0 +1,69 @@
...
+
+# Invalid URI
+write "abc://foo" [0x00]
```
jitsni: What's the end 0x00 (again, I see that gateway doing this way) ? If the length of address is 9, here we are sending 10 bytes.

specification/socks5plus/src/main/scripts/org/kaazing/specification/socks5plus/request/bind/connection.request.bind.with.invalid.uri.request.rpt
nowucca: To me this looks like field 4 of a Socks4a message with Kaazing uri address type.  The 0x00 0x00 0x00 0x09 is in fact saying "read a name until the null character is seen".  Probably not what we are expecting in a SOCKS5 message.  This could be a longer-term bug I introduced...?  Or we could say that the ATYP 'FE' needs to be processed like this by fiat.





https://en.m.wikipedia.org/wiki/SOCKS: "A server using protocol SOCKS4A must check the DSTIP in the request packet. If it represents address 0.0.0.x with nonzero x, the server must read in the domain name that the client sends in the packet."




—

PR #225, opened 2015-10-07
Fix #224 by ensuring that parent directory exists…

### LINKED ISSUE #224: file transport fails to create file when parent directory does not exist
The following script fails with an internal `NullPointerException` when the parent directory of the target file does not already exist.

```
connect file:target/k3po/echo.back.to.sender.dat
        option size 256
        option mode rw
connected

write option offset "0"
```

jfallows:
… before file is created when mode is rw

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannel.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannel.java
@@ -138,6 +138,9 @@ public final class FileChannel extends AbstractChannel<FileChannelConfig> {
                 break;
             case "rw":
                 mapMode = READ_WRITE;
+                // parent directory must exist to create file
+                File parentDir = location.getParentFile();
+                parentDir.mkdirs();
                 break;
             default:
                 throw new IllegalArgumentException(String.format("Unknown mode = %s for file = %s", mode, location));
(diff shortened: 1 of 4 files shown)


driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/file/FileChannel.java
```diff
@@ -138,6 +138,9 @@ private static MappedByteBuffer mapFile(URI fileAddress, String mode, long size)
                 break;
             case "rw":
                 mapMode = READ_WRITE;
+                // parent directory must exist to create file
+                File parentDir = location.getParentFile();
+                parentDir.mkdirs();
```
jitsni: I think we have two options. Either to create parent dir (as above) or throw exception if the parent dir doesn't exist.

Also, do we need to guard against NPE  that may occur due to other reasons (create may fail due to permissions etc).

jfallows: Throwing the exception doesn't get us a solution unless theres another way to create the parent directory, which would imply adding that capability to the K3PO scripting language (ugh!) or requiring the specification tests to have more than just `k3po.finish()` (double ugh!). :smile:

I think it would be good for you to add a different test case to verify the permissions behavior with a negative test.

jitsni: Sounds good. Good to merge.

PR #228, opened 2015-10-09
Normalize the path before passing to class loader to load resource

jitsni:
I am trying to use scripts from multiple locations(from specification jars and local file paths). Without normalization, classloader doesn't load the resources. This changeset fixes that issue.

```
private final K3poRule robot = new K3poRule().setScriptRoot("org/kaazing/specification/ws/framing");
@Specification({
       "echo.binary.payload.length.125/handshake.request.and.frame",
       "../../../../../com/kaazing/gateway/server/transport/socks/ws.socks.ws.data.exchange" })
    public void test() throws Exception {
```


jitsni:
@dpwspoon can you review it too

dpwspoon:
This is okay, I think I would prefer to be able to pass in script root as a prefix to the specification annotation.  So there would never be a need for ../../../

PR #240, opened 2015-10-23
Defer resolving accept and connect options until needed by accept or connect behavior

jfallows:
Resolve accept and connect options at the point of bind (accept) or connect, not during script parse or transformation.

Add workingDirectory to StartMojo, defaulted to ${basedir} to match surefire and failsafe, important for multi-module builds.


jitsni:
BarriersIT is failing, is that a known failure

PR #242, opened 2015-10-24
Using HttpClientCodec to deal with head responses with Content-Length

jitsni:


jitsni:
@jfallows I tried delegation to HttpClientCodec(almost worked, but couldn't delegate HttpResponseDecoder.replace() to HttpClientCodec's decoder). Hence used reflection at the end. One of the commits has the delegation classes.

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/message.format/head.response.must.not.have.content.though.may.have.content.length/response.rpt
```diff
@@ -5,7 +5,7 @@ connected
 read method "HEAD"
 read closed
 
-write status "100" "OK"
+write status "200" "OK"
 write header "Content-Length" "100"
 write close
```
jitsni: "write close" seems to close tcp stream. Is this expected ?

dpwspoon: Yes, sort of.  It was generally expected that when writing http level scripts that k3po could do anything valid at the tcp layer.  Closing the connection here is valid so its okay.  We have never implemented semantics to control the tcp layer, from a k3po script that is written in the http layer

PR #245, opened 2015-10-29
HTTP proxy sending 504 status code

jitsni:


jitsni:
@msalavastru good work. I am creating a new PR, see if you have any comments

jitsni:
One test that is missing: when server takes too much time to respond, then proxy need to send 504 to client. May be it can be addressed when we do DOS story.

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/server.error/proxy.should.return.504.response.when.server.disconnects/proxy.rpt
```diff
@@ -0,0 +1,33 @@
...
+accepted
+connected
+
+read method "GET"
+read header "Host" "localhost:8080"
+read notify GOT_REQUEST
+read closed
+
+write await GOT_RESPONSE
+write status "504" "OK"
```
msalavastru: This should be **write status "504" "Gateway Timeout"**

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/server.error/proxy.should.return.504.response.when.server.disconnects/proxy.rpt
```diff
@@ -0,0 +1,33 @@
+#
+# client <--> proxy
+#
+accept http://localhost:8080/index.html
+accepted
+connected
+
+read method "GET"
```
msalavastru: Should add **read version "HTTP/1.1"**

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/server.error/proxy.should.return.504.response.when.server.disconnects/proxy.rpt
jitsni: Created a PR for this.

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/server.error/proxy.should.return.504.response.when.server.disconnects/proxy.rpt
jitsni: Not that important to check version in this test, but included it in the PR

PR #248, opened 2015-10-30
K3po issue 222 close vs disconnected

cmebarrow:
The issue was occurring because of the reliance of CompositeChannelFuture on its listener being fired on every kid future before it reports completion. This messes up when certain events like CloseHandler actually cause other events (disconnected, etc.) to occur on stack before all the future listeners have had a chance to be executed. The fix is to avoid trying to keep count of unnotified, succeeded, etc kids and instead scan the kids each time to get an up-to-date picture of whether all have completed or not.

Prior to this fixed the test case closed.vs.disconnected.issue.rpt was failing with expected script closed (on the first stream) and actual disconnected.

NOTE: NOT READY FOR MERGE YET!!!


cmebarrow:
This fix has not fixed the issue! The new test (in driver/BehaviorIT) always passes with the fix on my new work laptop (and always failed before). However on my old work laptop it fails about 1 run out of every two.

cmebarrow:
I have now made a fix which allows build to succeed on my old and new work laptops. However it still fails on Travis, in specification/bbosh.

cmebarrow:
Closing because we need a better fix which will be done in a different branch. This fixes not fully resolving the underlying issue which is command handlers (like CloseHandler) causing upstream events to occur before the current upstream event has finished processing. The new fix will attempt to schedule command handler actions onto the IO worker task queue so the processing of the current upstream event can complete before the command is executed.

PR #254, opened 2015-11-04
Support accept notify to be used in conjuction with connect await

jfallows:


dpwspoon:
LGTM, would like to have a unit test in driver for this functionality...

Issue #262, opened 2015-11-13
specification/bbosh build failed in test BBoshSpecificationIT, probably out of order connects


cmebarrow:
I got a failure in shouldEchoThenClosedViaStreaming in a Travis build. Output is attached:
[build44_issue_261_fix_branch_bbosh.fail.txt](https://files.zenhub.io/564549d18294d8de6364c878)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

This is probably due to an issue with the test client script (/specification.bbosh/src/main/scripts/org/kaazing/specification/bbosh/streaming/connect.echo.then.closed.rpt). It needs to do connect await to delay the second http request, instead of just doing write await.

cmebarrow:
Fixed in PR #263

Issue #260, opened 2015-11-10
http specification tests ConnectionManagementIT, TransferCodingsIT failing with read... instead of closed


cmebarrow:
The failures indicate data was read unexpectedly before the closed event came in:
```
Failed tests:
  ConnectionManagementIT.serverMustCloseConnectionAfterResponseWithConnectionClose Specified behavior did not match expected:<...write "\r\n"

close
[closed]

> but was:<...write "\r\n"

close
[read ...]

>
  TransferCodingsIT.shouldProcessRequestTransferEncodingChunked Specified behavior did not match expected:<... 200 OK\r\n"

close
[closed]

accept http://loca...> but was:<... 200 OK\r\n"

close
[read ...]

accept http://loca...>
```

cmebarrow:
For the first failure at the issue is that script `specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/server.must.close.connection.after.response.with.connection.close/response.rpt` fails to read all the data written by the request script. It is a miracle this test ever passed!

cmebarrow:
I have a fix ready

cmebarrow:
Fixed in PR #263

Issue #259, opened 2015-11-09
wse specification test DownstreamIT sometimes fails because HTTP requests are executed out of order


cmebarrow:
This is a very similar issue to #258. Test failures may occur because k3po does not by default guarantee the order of execution of HTTP request/response pairs. The scripts need to be changed to use barriers to control the ordering.

cmebarrow:
Fixed in PR #263

Issue #256, opened 2015-11-05
http script with closed causes test to hang or time out


cmebarrow:
The following script (which I will be adding to driver HttpIT shortly) causes the test to hang (or time out if there is a timeout rule in the test):
```
# client
connect http://localhost:8080/index.html
connected

write method "GET"
write version "HTTP/1.1"
write header host
write close

# the following causes the test to hang (or be timed out by a timeout rule)
closed

# server
accept tcp://localhost:8080
accepted
connected

read "GET /index.html HTTP/1.1\r\n"
read "Host: localhost:8080\r\n"
read "\r\n"

close
closed
```
This seems to be the underlying cause of failures of HTTP specification test ResponseStatusCodesIT.proxyShouldReturn504ResponseWhenServerDisconnects on Travis.

cmebarrow:
Strangely if the read commands are removed the test passes if it times out, even though driver debug output shows there are failures.

jfallows:
This looks like a _reset_ scenario where the TCP connection is torn down before any response is given to the client. 

Shouldn't this give an exception on the pipeline and fail the test?

Btw, how could the server provide a 504 response if the server has already closed the underlying TCP connection?

cmebarrow:
The scenario for ResponseStatusCodesIT.proxyShouldReturn504ResponseWhenServerDisconnects is more elaborate than the test case I gave in the above description. The proxy sends back response 504 because the server it connected to seven is the connection before responding.

PR #263, opened 2015-11-14
Script corrections to reduce build failures, especially on Travis, plus some diagnostic improvements

cmebarrow:
- (#262) Fixed two scripts for BBoshSpecificationIT to use connect await to order http requests.
- Added test case for k3po issue #237.
- (#260) fixed some of the http specification scripts for rfc7230 tests ConnectionManagementIT and TransferCodingsIT to read all incoming data before doing close closed, and so server sends the final \r\n so it's a valid http response, to avoid sporadic test failures.
- (#259) fix script for DownstreamIT to use barriers to guarantee the HTTP requests are executed in the desired order (checked all specification/wse for this issue, only one needed fixing).
- Moved driver test case for #222 into new TcpIT test. Added test cases for #256 (in HttpIT) and #128 (in TcpIT). Fixed script delayed.http.connect.rpt to add missing \r\n.
- Added barriers to all muliple request http specification scripts for StatusCodeDefinitionIT because k3po does not otherwise guarantee the ordering of successive HTTP request/response pairs (sporadic test failures were occurring because of this)
- Removed "closed" from proxy.rpt for http specification test case ResponseStatusCodesIT.proxyShouldReturn504ResponseWhenServerDisconnects to work around issue #256.
- Add test case for issue #222 (close vs disconnected), disabled for now.
- Added test execution log messages to TcpIt and HttpIT to get more info for test failures.
- Add more info (pipelineFuture) to debug message in ExecutionHandler 
- Changed two debug log messages (on test failure) to info level in ExecutionHandler and ControlServerHandler to try to get more info for test failures.


cmebarrow:
The Travis build failed because of known bug #202

driver/src/test/java/org/kaazing/k3po/driver/internal/http/HttpIT.java
```diff
@@ -219,6 +233,14 @@ public void shouldAcceptMultipleHttpOnSameTcp() throws Exception {
     }
 
     @Test
+    @Ignore("kepo#256")
```
dpwspoon: typo

cmebarrow: fixed

PR #266, opened 2015-11-18
Simplify debug output...

jfallows:
...for improved readability when debugging.


specification/tcp/src/test/java/org/kaazing/specification/tcp/rfc793/TcpIT.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
```
dpwspoon: ping @NicoletaOita this will require updates to consuming tests

specification/tcp/src/main/scripts/org/kaazing/specification/tcp/rfc793/client.sent.data/server.rpt
```diff
@@ -1,5 +1,5 @@
 accept tcp://localhost:8080
+       notify BOUND
```
dpwspoon: I think this has to do more with our conversation about perhaps removing the need for k3po.start() and k3po.finish() and less to do with the purpose of this PR ("Simplify Debug Output").  I'm reluctant to merge this part of the PR until we agree on the end game of how this will work and backwards compatibility.  I'd rather move quickly towards getting a final design and implementation for that done and consumed everywhere, rather then implement it piecemeal.  I've open https://github.com/k3po/k3po/issues/267 to track that conversation.

PR #279, opened 2015-12-28
Removed 0x7f from randomBytesIncludingNumberOfEscapedBytes as it make…

dpwspoon:
…s websocket frame length undeterministic as it goes over utf 8.  We will test escaped bytes explicitly elsewhere

### DIFF
--- a/specification/wse/src/main/java/org/kaazing/specification/wse/internal/Functions.java
+++ b/specification/wse/src/main/java/org/kaazing/specification/wse/internal/Functions.java
@@ -48,7 +48,13 @@ public final class Functions {
     @Function
     public static byte[] randomBytesIncludingNumberOfEscapedBytes(int length, int numberOfEscapedBytesToInclude) {
         byte[] bytes = new byte[length];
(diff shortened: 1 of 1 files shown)


specification/wse/src/main/java/org/kaazing/specification/wse/internal/Functions.java
```diff
@@ -48,7 +48,13 @@
     @Function
     public static byte[] randomBytesIncludingNumberOfEscapedBytes(int length, int numberOfEscapedBytesToInclude) {
         byte[] bytes = new byte[length];
-        byte[] escapedBytes = {0b00000000, 0b00001101, 0b00001010, 0b01111111};
+        byte[] escapedBytes = {0b00000000, 0b00001101, 0b00001010};
+        // escluding 0b01111111
```
cmebarrow: excluding

cmebarrow:
"We will test escaped bytes explicitly elsewhere" Where?

dpwspoon:
This is not a correct fix as 0b01111111 should be valid single utf-8 byte

PR #280, opened 2015-12-29
Added interpretation to wse spec on correct behavior. 

dpwspoon:
Confirmed that is how the test is working in kaazing gateway 4.x.

@cmebarrow Having @jfallows review but thought you might be interested.


specification/wse/src/main/scripts/org/kaazing/specification/wse/data/binary.as.escaped.text/echo.payload.length.127/request.rpt
```diff
@@ -1,5 +1,5 @@
 property sequence ${wse:randomInt(100)}
```
dpwspoon: This test is ignored and will be fixed assuming the interpretation of the spec above is correct

specification/wse/SPEC.md
```diff
@@ -642,14 +642,19 @@ interpret character code zero as end-of-response, truncating any body content fo
 canonicalizes carriage return and linefeed characters as all carriage returns.  Conversely, IE9 strict document mode 
 canonicalizes carriage return and linefeed characters as all linefeeds.
 
-Therefore, the server MUST escape these characters must as follows:
+Therefore, the client and server MUST escape these characters must as follows (in both upstream and downstream):
```
cmebarrow: Need to remove the second "must"

specification/wse/src/main/scripts/org/kaazing/specification/wse/data/binary.as.escaped.text/echo.escaped.characters/request.rpt
```diff
@@ -0,0 +1,55 @@
...
+write header host
+write header "X-Sequence-No" ${wse:asString(sequence + 1)}
+write header content-length
+write close
+
+read status "200" /.+/
+read header "Content-Type" "text/plain;charset=windows-1252"
+read header "Connection" "close"
+
+read [0x80 0x04 0x7f 0x30 0x7f 0x72 0x7f 0x6e 0x7f 0x7f]
```
cmebarrow: Why isn't this 0xC2 0x80?

dpwspoon: ["windows-1252 is a superset of IEC_8859-1"](https://en.wikipedia.org/wiki/Windows-1252).  From [IEC_8859-1](https://en.wikipedia.org/wiki/C0_and_C1_control_codes) The popular Windows-1252 character set adds all the missing characters provided by ISO/IEC 8859-15, plus a number of typographic symbols, by replacing the rarely used C1 controls in the range 128 to 159 (hex 80 to 9F).  Thus hex 80 is a valid character in windows-1252
which is padding, or sometimes a euro sign in browsers: http://www.floyd.ch/?p=12.  

Here is a [unicode mapping of 1252](http://www.unicode.org/Public/MAPPINGS/VENDORS/MICSFT/WINDOWS/CP1252.TXT),  0x80 is the euro sign

dpwspoon: Perhaps I'll add that to the spec.  The 0xC2 0x80 that is below is in the charset=utf-8 section where 0x80 needs extended bytes to represent it

cmebarrow: Yes please do add this to the spec

dpwspoon:
I've confirmed with John that the spec changes are correct

cmebarrow:
@dpwspoon Since John is OK with the spec changes do you want me to go ahead and merge this? I have reviewed it and it all looks good to me. It's nice to have this all clarified, great work!

dpwspoon:
@cmebarrow, that sounds good.  I was planning to go back and update the one test that is currently ignored. But I'm fine with you taking that over if you have time or it falls under your current sprint tasks?

cmebarrow:
@dpwspoon Could you resolve the conflicts then I can merge and update that test as part of my wse spec test integration work

dpwspoon:
@cmebarrow I've updated.

Issue #261, opened 2015-11-12
k3po build failed in driver module due to failed test K3poIT.connectFinishedWithBarriers


cmebarrow:
I have only so far seen this failure once, in this Travis build, so it's obviously due to a rare race condition. The build output is attached:
[build39_diagnostics_with_buffered_logger.fail.in.driver.k3poIT.txt](https://files.zenhub.io/56450a198294d8de6364c5de)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

The build output shows the failure occurs because the FINISHED bbosh message was received before the NOTIFIED message.

cmebarrow:
This bug can easily be reproduced in a debugger. Set breakpoints in ControlServerHandler as follows:
- in the listener set within method writeNotifiedOnBarrier
- in the listener set within method whenAbortedOrFinished
Now run the test in debug mode and do as follows:
- When the former breakpoint is first hit it will be for barrier "NOTIFY_BARRIER". Hit continue.
- When the same breakpoint is again (for barrier "AWAIT_BARRIER") leave the thread blocked
- When the second breakpoint is hit, hit continue to allow the FINISHED method to be written out
- Now hit continue for the thread in which the first breakpoint was hit, allowing the NOTIFIED message to be written out. Since it is written after FINISHED the test fails.

PR #284, opened 2016-01-08
Make sure BBOSH FINISH is sent only after sending all NOTIFIED messages 

cmebarrow:
The purpose is to fix bug #261. The change is to ControlServiceHandler.


dpwspoon:
LGTM

PR #290, opened 2016-01-13
Wse: improve spec and tests

cmebarrow:
This PR seeks to improve and complete the wse specification (SPEC.md) and tests. 

The test work is not yet complete, but I would appreciate feedback on the SPEC.md changes.

The main changes to SPEC.md are:

* [Data Frames](https://github.com/cmebarrow/k3po/blob/wse_improve_coverage/specification/wse/SPEC.md#data-frames) section: give details about text frames
* Rename Binary as Text section to [Text Encoding](https://github.com/cmebarrow/k3po/blob/wse_improve_coverage/specification/wse/SPEC.md#text-encoding), change it to cover both /ctm and /ct, and give more details about how the upstream payload is UTF-8 encoded (everything is encoded, not just binary frames)
* Rename Binary as Escaped Text section to [Text Escaped Encoding](https://github.com/cmebarrow/k3po/blob/wse_improve_coverage/specification/wse/SPEC.md#text-escaped-encoding), change it to cover both /ctem and /cte.


chao-sun-kaazing:
yes. SPEC is better now

specification/wse/SPEC.md
```diff
@@ -317,26 +319,22 @@ Connection: close
...
 
-The downstream response MAY use `Connection: close` or `Transfer-Encoding: chunked` to provide a continuously streaming response
-to the client.
+The downstream response MAY use `Connection: close` or `Transfer-Encoding: chunked` to provide a continuously streaming response to the client.
 
 See [Buffering Proxies](#buffering-proxies) for further client requirements when attaching the downstream.
 
-If the downstream HTTP response transfer is complete, or else complete but does not end in a `RECONNECT` command frame, then the 
-client should consider this as unexpected connection loss for the emulated WebSocket connection.
+If the downstream HTTP response transfer is complete, or else complete but does not end in a `RECONNECT` command frame, then the client should consider this as unexpected connection loss for the emulated WebSocket connection.
```
sanjay-saxena: It would be helpful if the language here can be simplified/tightened. Specifically, the `or else` part throws me off every single time.

specification/wse/SPEC.md
```diff
@@ -426,12 +422,25 @@ The client requirements for data frame syntax are defined by
 The server requirements for data frame syntax are defined by 
 [Draft-76, Section 5.3](http://tools.ietf.org/html/draft-hixie-thewebsocketprotocol-76#section-5.3).
 
+Binary frames are represented using frame type 0x80.
+
+Text frames may be represented in one of two formats:
+
+* Delimited: the frame type byte is 0x00, with the 0xFF delimiter marking end of data
+* Specified length: the frame type is 0x81, which, according to the rules of the above referenced sections of [Draft-76](http://tools.ietf.org/html/draft-hixie-thewebsocketprotocol-76), implies that the length is specified at the start of frame.
```
sanjay-saxena: Note that `0x81` is NOT defined by Draft-76. It is a Kaazing-specific frame type indicating that the length is specified at the start of the text frame.

specification/wse/SPEC.md
```diff
@@ -426,12 +422,25 @@ The client requirements for data frame syntax are defined by
...
+
+If the server receives a text frame that contains invalid UTF-8, the server MUST generate an HTTP response with a `400 Bad Request` status code and fail the WSE connection.
+
 ## Command Frames
 
 The frames used by an emulated WebSocket connection for upstream and downstream data transfer extend those defined by 
 [Draft-76](http://tools.ietf.org/html/draft-hixie-thewebsocketprotocol-76), by adding a command frame.
 
-The text-based command frame has frame type `0x01`, which masks to `0x00` indicating text content.
+The text-based command frame has frame type `0x01`, which does not have the leading bit set so masks to `0x00` indicating text content.
```
sanjay-saxena: Thanks for making that explicitly clear!

specification/wse/SPEC.md
```diff
@@ -635,7 +644,37 @@ because each character is represented by a single byte and exactly 256 distinct
...
+Rather than escaping the `\0` (NUL) byte-as-character-code, the client MAY use character code `256` to represent zero, so the
+client MAY send a `\u0100` character for each `\u0000` character.
+
+The server MUST decode the binary-as-mixed-text upstream such that each character codepoint is computed `mod 0x0100` to determine each originally intended byte value.
+
+The upstream HTTP request MUST have content type `text/plain;charset=utf-8`. The upstream request body MUST contain one or more valid emulated WebSocket frames encoded into valid UTF-8 by transforming each byte as follows:
+
+* If the leading (most significant) bit is set, then the byte MUST the encoded as two bytes, in the form `1100 00?? 10?? ????` where `?` represents one bit of the original byte
+* If the leading bit is not set, then the byte is encoded as is
+* If the byte value is 0x00 it MAY be encoded as 0xC4 0x80 (instead of 0x00)
```
sanjay-saxena: Perhaps, I am mistaken. But, I want to make sure that we got this right. Will `0x00` get encoded as `0xC0 0x80` based on the encoding scheme that you have specified?

sanjay-saxena: Looks good, @cmebarrow!

cmebarrow: I think it's 0xC4 0x80 because that is
`‭‭1100 0100 1000 0000‬'
which is UTF-8 codepoint binary 001 0000 0000 (last 5 bites of first byte plus last 6 of second) which is denary 256. See https://en.wikipedia.org/wiki/UTF-8. Does that make sense or am I misunderstanding something?

sanjay-saxena: Maybe, I got it wrong. If it should be `0xC4 0x80`, then should the part where you say `in the form 1100 00?? 10?? ????` should be changed to ` in the form 1100 01?? 10?? ????` ?

specification/wse/SPEC.md
cmebarrow: Correct, but 0x81 is compatible with Draft-76 because Draft-76 only talks about the leading bit of the frame type byte and poses no constraints on the remaining bits or the meaning and interpretation of the frame payload. If set then the message is specified length format.

specification/wse/SPEC.md
cmebarrow: I have removed ", or else complete" so it now reads "If the downstream HTTP response transfer is complete but does not end in a `RECONNECT` command frame, then the client should consider..."

specification/wse/SPEC.md
cmebarrow: Added a sentence stating this

specification/wse/SPEC.md
cmebarrow: No, because the two cases are different. The sentence containing "in the form 1100..." Is talking about how to encode **8-bit** values (bytes) which have the leading bit set. In the case of 0xC4 0x80 we are talking about how to represent the **9-bit** value 1 0000 0000.

PR #291, opened 2016-01-14
SSE spec tests

sanjay-saxena:


dpwspoon:
Looks good @sanjay-saxena.  Can we add a SPEC.md with link to relevant RFC or equivalent.  I always look at https://www.w3.org/TR/2009/WD-eventsource-20090423/, but I'm not sure that is the best source or not.

I didn't see tests about closing the connections, or re opening the connections in the case of a retry/errors.  It's not 100% clear to me how a reconnection happens.  Though it seems like we should have a test for it.  Also the following section doesn't seem to be represented in this spec though arguably they might be covered in the http spec.  If so, maybe make a note of it in the SPEC.md, or consider adding tests for it.  Also, its not clear to me what the user agent should do if the server sends bad behavior.  I believe browsers treat this as a network error and close the connection.   If that is in the Spec we should add a test for it.

> If such a resource (with the correct MIME type) completes loading (i.e. the entire HTTP response body is received or the connection itself closes), the user agent must reset the connection. This doesn't apply for the error cases that are listed below.
>
HTTP 200 OK responses that have a Content-Type other than text/event-stream (or some other supported type) must cause the user agent to fail the connection.
>
HTTP 204 No Content, and 205 Reset Content responses are equivalent to 200 OK responses with the right MIME type but no content, and thus must reset the connection.
>
Other HTTP response codes in the 2xx range must similarly reset the connection. They are, however, likely to indicate an error has occurred somewhere and may cause the user agent to emit a warning.
>
HTTP 301 Moved Permanently responses must cause the user agent to reconnect using the new server specified URL instead of the previously specified URL for all subsequent requests for this event source. (It doesn't affect other EventSource objects with the same URL unless they also receive 301 responses, and it doesn't affect future sessions, e.g. if the page is reloaded.)
>
HTTP 302 Found, 303 See Other, and 307 Temporary Redirect responses must cause the user agent to connect to the new server-specified URL, but if the user agent needs to again request the resource at a later point, it must return to the previously specified URL for this event source.
>
HTTP 305 Use Proxy, HTTP 401 Unauthorized, and 407 Proxy Authentication Required should be treated transparently as for any other subresource.
>
Any other HTTP response code not listed here or network error (e.g. DNS errors) must cause the user agent to fail the connection.

specification/sse/src/main/scripts/org/kaazing/specification/sse/unnamed.event/multiple.events/response.rpt
```diff
@@ -0,0 +1,43 @@
...
+read "\r\n"
+
+write "HTTP/1.1 200 OK\r\n"
+write "Content-Type: text/event-stream\r\n"
+write "Cache-Control: no-cache\r\n"
+write "Connection: keep-alive\r\n"
+write "Date: " ${date} "\r\n"
+write "\r\n"
+
+write "data:Message for unnamed server-sent event!\n"
```
dpwspoon: Consider distinguishing the first and last message here.  Currently they both are "Message for unnamed server-sent event!"

sanjay-saxena:
I have added README.md. And, a test for invalid `Content-Type` which may not be that useful as there isn't much that happens until the connection is established. Also, note that should (this)[https://www.w3.org/TR/eventsource/] is the latest spec.

There isn't any notion of closing a connection with respect to SSE. If there is no activity, browsers close the connection after 5minutes. Similarly, servers such as `nginx ` provide a way to specify the inactivity period after which a connection is closed.

As part of the event stream, the publisher of the events can send a named event which the client application can check to gracefully invoke `EventSource.close()` to close a connection on the client. The connection will stay for a while on the server and then gets closed eventually.

If `EventSource.close()` is not invoked and the underlying connection was closed due to some other reason, then the browser will keep reissuing `GET` requests to reconnect using the `retry` interval. These `GET` requests can include the `Last-Event-Id` header to catch up with the stream. I already have a test for this scenario.

dpwspoon:
Talked to Sanjay, will add the invalid utf-8 test which will be useful for client sides

PR #293, opened 2016-01-16
Httpxe specification tests

jitsni:
Gateway implements the specification tests in
https://github.com/kaazing/gateway/pull/399


specification/httpxe/src/main/scripts/org/kaazing/specification/httpxe/responses/wrapped.101.response.in.200/request.rpt
```diff
@@ -19,11 +19,17 @@ connected
...
+write header host
+write header "Content-Type" "application/x-message-http"
+write header "X-Next-Protocol" "httpxe/1.1"
+write header content-length
+write "GET /path HTTP/1.1\r\n"
+      "Content-Length: 0\r\n"
+      "\r\n"
+write close
 
 read status "200" "OK"
```
jitsni: From the specification, 1xx status code will be in the enveloped response. Here it is 200. Should this be changed ?

specification/httpxe/src/main/scripts/org/kaazing/specification/httpxe/responses/wrapped.501.response.in.200/request.rpt
```diff
@@ -19,12 +19,17 @@ connected
...
+write header host
+write header "Content-Type" "application/x-message-http"
+write header "X-Next-Protocol" "httpxe/1.1"
+write header content-length
+write "GET /path HTTP/1.1\r\n"
+      "Content-Length: 0\r\n"
+      "\r\n"
+write close
 
 read status "200" "OK"
```
jitsni: Specification says 5xx status codes should be in enveloped response. Here it is 200. Should this be changed ?

jfallows:
Looks good, @jitsni.
Make sure to use a regex to read the status description text since they are not required to be the canonical text.

Issue #299, opened 2016-01-26
HTTPX: x-kaazing-handshake protocol - Empty WebSocket binary frame marks the end of extended handshake response


sanjay-saxena:
`x-kaazing-handshake` protocol as defined in [HTTPX specification](https://github.com/sanjay-saxena/k3po/blob/develop/specification/httpx/SPEC.md) must include an empty WebSocket binary frame at the end of extended handshake response. The empty WebSocket binary frame from the server looks as shown below:

`0x82 0x00`

Since the server may break the extended handshake response into multiple WebSocket frames, an empty WebSocket binary frame is used to indicate the end of the extended handshake response.

This means we have to update the specification as well as the existing spec scripts to include the empty WebSocket binary frame.

cmebarrow:
@jfallows WDYT? Do you think this is the right approach?

sanjay-saxena:
Here is the [PR](https://github.com/k3po/k3po/pull/302).

jfallows:
As far as I remember, the empty WebSocket frame was a bug in the original implementation and should not be specified if not functionally needed.

Can we remove the requirement for the empty binary frame from the spec?

sanjay-saxena:
Since the extended handshake response can be split across multiple WebSocket binary frames, currently the Gateway indicates the end of extended handshake response to the clients using an empty WebSocket frame. For example, here is how the extended handshake response looks when Gateway needs to send sending a `401` to the client:

```
# Extended handshake response - WebSocket Frame1
0x82 0x41
"HTTP/1.1 200 OK\r\n"
"Content-Type: text/html\r\n"
"Content-Length: 194\r\n"
"\r\n"

# Extended handshake response - WebSocket Frame2
0x82 0x7e 0x00 0x82
"HTTP/1.1 401 Unauthorized\r\n"
"WWW-Authenticate: Application Basic realm="
0x22
"Kaazing WebSocket Gateway Demo"
0x22
"\r\n"
"Content-Type: text/html\r\n"
"\r\n"

# Extended handshake response - WebSocket Frame3
0x82 0x40
"<html><head></head><body><h1>401 Unauthorized</h1></body></html>"

# Extended handshake response - WebSocket Frame4
0x82 0x00
```

jfallows:
Thanks for the very clear explanation @sanjay-saxena. :smile: 

Wouldn't it make sense just to follow the `Content-Length` header value to determine when all bytes representing the response have been processed?

[...]

PR #302, opened 2016-01-27
Updated the port numbers and the paths in the scripts to match the ones used in client itests

sanjay-saxena:
This would make it easier to run the client itests against the scripts, eventually.


cmebarrow:
@sanjay-saxena Did you run this by @jfallows ?

sanjay-saxena:
This will not be merged as-is because @jfallows likes to move the empty frames for backward compatibility out of the spec. Once we have a consensus, we will make the changes accordingly. Look at #299 for the on-going conversation.

sanjay-saxena:
@jfallows - Please hold on the merge as this is not as simple as we originally thought. I will update the [ticket](https://github.com/kaazing/roadmap/issues/669) with details. Based on my findings, we will need more changes to the specification scripts and SPEC.md.

specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
```diff
@@ -15,7 +15,7 @@
 #
 
 property writeMask ${httpx:randomBytes(4)}
```
dpwspoon: I think this was a mistake.  We should use standard same ports and paths ("8080" and "path"), just like all the other specification tests.  @sanjay-saxena can you back out this change, and switch to 8080 and path

Issue #297, opened 2016-01-25
Barrier blocked http is not always fired.


dpwspoon:
The following script is sometimes not firing the bottom http request

 ```
+#
 +# Copyright (c) 2007-2015, Kaazing Corporation. All rights reserved.
 +#
 +property writeMask ${ws:randomBytes(4)}
 +connect http://localhost:8001/echo?.kl=Y
 +connected
 +
 +write method "GET"
 +write version "HTTP/1.1"
 +write header host
 +write header "Upgrade" "websocket"
 +write header "Connection" "Upgrade"
 +write header "Sec-WebSocket-Key" "dGhlIHNhbXBsZSBub25jZQ=="
 +write header "Sec-WebSocket-Version" "13"
 +write header "Sec-WebSocket-Protocol" "x-kaazing-handshake"
 +
 +read status "101" /.+/
 +read version "HTTP/1.1"
 +read header "Upgrade" "websocket"
 +read header "Connection" "Upgrade"
 +read header "Sec-WebSocket-Accept" "s3pPLMBiTxaQ9kYGzzhZRbK+xOo="
 +read header "Sec-WebSocket-Protocol" "x-kaazing-handshake"
 +read header "Server" /.+/
 +read header "Date" /.+/
 +read [0x82 0x04]
 +read [0xef 0x83 0xbf 0x4e]
 +
 +write [0x82 0xC6] ${writeMask}
 +write option mask ${writeMask}
 +write "GET /echo HTTP/1.1\r\n"
 +write "WebSocket-Protocol: \r\n"
 +write "Sec-WebSocket-Protocol: \r\n"
 +write "\r\n"
 +write option mask [0x00 0x00 0x00 0x00]
 +
 +# 101 response
 +read [0x82 0x7e 0x00 0xc9]
 +read "HTTP/1.1 101 Web Socket Protocol Handshake\r\n"
 +read "Connection: Upgrade\r\n"
 +read "Date: " /.+/ "\r\n"
 +read "Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n"
 +read "Server: Kaazing Gateway\r\n"
 +read "Upgrade: websocket\r\n"
 +read "\r\n"
 +
 +read [0x82 0x00]
 +
 +read [0x82 0x23 0xef 0x83 0xbf 0x52] 
 +read "ws://localhost:8001/echo1?.kl=Y"
 +read [0x88 0x02 0x03 0xe8]
 +
[...]

Issue #202, opened 2015-06-26
FAILURE! - in org.kaazing.specification.http.rfc7230.MessageFormatIT


a-zuckut:
```code
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.kaazing.specification.http.rfc7231.RequestHeaderFieldsIT
Tests run: 3, Failures: 0, Errors: 0, Skipped: 3, Time elapsed: 0.013 sec - in org.kaazing.specification.http.rfc7231.RequestHeaderFieldsIT
Running org.kaazing.specification.http.rfc7231.RequestMethodsIT
Tests run: 4, Failures: 0, Errors: 0, Skipped: 4, Time elapsed: 0.006 sec - in org.kaazing.specification.http.rfc7231.RequestMethodsIT
Running org.kaazing.specification.http.rfc7235.HeaderFieldDefinitionsIT
Tests run: 7, Failures: 0, Errors: 0, Skipped: 7, Time elapsed: 0.012 sec - in org.kaazing.specification.http.rfc7235.HeaderFieldDefinitionsIT
Running org.kaazing.specification.http.rfc7235.AuthIT
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.43 sec - in org.kaazing.specification.http.rfc7235.AuthIT
Running org.kaazing.specification.http.rfc7230.MessageRoutingIT
line 5:29 token recognition error at: '"\r\n\"\n'
line 6:31 token recognition error at: '"\r\n\"\n'
line 16:28 token recognition error at: '"\r\n\"\n'
line 17:30 token recognition error at: '"\r\n\"\n'
line 5:31 token recognition error at: '"\r\n\"\n'
line 15:30 token recognition error at: '"\r\n\"\n'
Tests run: 17, Failures: 0, Errors: 0, Skipped: 11, Time elapsed: 0.454 sec - in org.kaazing.specification.http.rfc7230.MessageRoutingIT
Running org.kaazing.specification.http.rfc7230.ConnectionManagementIT
Tests run: 12, Failures: 0, Errors: 0, Skipped: 3, Time elapsed: 0.709 sec - in org.kaazing.specification.http.rfc7230.ConnectionManagementIT
Running org.kaazing.specification.http.rfc7230.ArchitectureIT
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.588 sec - in org.kaazing.specification.http.rfc7230.ArchitectureIT
[...]

Issue #233, opened 2015-10-14
Specification.WSE: add upstreaming support using chunking


cmebarrow:
Currently the [WSE specification](https://github.com/k3po/k3po/blob/develop/specification/wse/SPEC.md) does not allow a viable means of streaming upstream data, because it stipulates "If the upstream HTTP request transfer is either incomplete or else complete but does not end in a RECONNECT command frame, then the server should consider this as unexpected connection loss for the emulated WebSocket connection.".

Currently our WsebConnector implementation achieves upstreaming by setting a very long content length on the upstream request and then continuously writing more data to the body. This does not allow for a clean way to close the connection, and is probably at least violating the spirit of the WSE specification, if not the letter. 

We should update the WSE specification to state that chunking may be used on the upstream when streaming is desired. That way a client (like WsebConnector) can choose continuously to send upstream data on a single HTTP request and network connection rather than having to issue multiple separate HTTP requests.

Issue #136, opened 2015-02-18
Http Sematics to support chunked encoding extensions and trailer


dpwspoon:
As per [spec](https://tools.ietf.org/html/rfc7230#section-4.1), transfer encoding chunk can have extensions and a trailer.  As of now there is no way to set that in http scripts.

>
>4.1.  Chunked Transfer Coding
>
>   The chunked transfer coding wraps the payload body in order to
>   transfer it as a series of chunks, each with its own size indicator,
>   followed by an OPTIONAL trailer containing header fields.  Chunked
>   enables content streams of unknown size to be transferred as a
>   sequence of length-delimited buffers, which enables the sender to
>   retain connection persistence and the recipient to know when it has
>   received the entire message.
>
>     chunked-body   = *chunk
>                      last-chunk
>                      trailer-part
>                      CRLF
>
>     chunk          = chunk-size [ chunk-ext ] CRLF
>                      chunk-data CRLF
>     chunk-size     = 1*HEXDIG
>     last-chunk     = 1*("0") [ chunk-ext ] CRLF
>
>     chunk-data     = 1*OCTET ; a sequence of chunk-size octets
>
>   The chunk-size field is a string of hex digits indicating the size of
>   the chunk-data in octets.  The chunked transfer coding is complete
>   when a chunk with a chunk-size of zero is received, possibly followed
>   by a trailer, and finally terminated by an empty line.
>
>   A recipient MUST be able to parse and decode the chunked transfer
>   coding.
>
>4.1.1.  Chunk Extensions
>
>   The chunked encoding allows each chunk to include zero or more chunk
>   extensions, immediately following the chunk-size, for the sake of
>   supplying per-chunk metadata (such as a signature or hash),
>   mid-message control information, or randomization of message body
>   size.
>
>     chunk-ext      = *( ";" chunk-ext-name [ "=" chunk-ext-val ] )
>
>     chunk-ext-name = token
>     chunk-ext-val  = token / quoted-string
>
>   The chunked encoding is specific to each connection and is likely to
>   be removed or recoded by each recipient (including intermediaries)
>   before any higher-level application would have a chance to inspect
>   the extensions.  Hence, use of chunk extensions is generally limited
>
>
>
>
>Fielding & Reschke           Standards Track                   [Page 36]
> 
>RFC 7230           HTTP/1.1 Message Syntax and Routing         June 2014
>
>
>   to specialized HTTP services such as "long polling" (where client and
>   server can have shared expectations regarding the use of chunk
[...]

PR #303, opened 2016-02-02
Adds abort and aborted command

dpwspoon:
This is ready for review.  I will also work on implementing the tests, perhaps in a different PR depending on when this gets merged

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
@@ -43,7 +43,6 @@ import java.util.NavigableMap;
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.channel.ChannelConfig;
-import org.jboss.netty.channel.ChannelException;
 import org.jboss.netty.channel.ChannelFactory;
 import org.jboss.netty.channel.ChannelFuture;
 import org.jboss.netty.channel.ChannelHandlerContext;
@@ -108,9 +107,7 @@ public class HttpChildChannelSource extends HttpChannelHandler {
             case CONTENT_COMPLETE:
                 break;
             default:
-                ChannelException exception = new ChannelException("Channel closed unexpectedly");
-                exception.fillInStackTrace();
-                fireExceptionCaught(httpChildChannel, exception);
+                Channels.fireChannelAborted(httpChildChannel);
                 break;
             }
 
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
@@ -62,6 +62,7 @@ import org.jboss.netty.handler.codec.http.QueryStringDecoder;
 import org.jboss.netty.handler.codec.http.QueryStringEncoder;
 import org.kaazing.k3po.driver.internal.netty.bootstrap.http.HttpChildChannel.HttpReadState;
 import org.kaazing.k3po.driver.internal.netty.channel.ChannelAddress;
+import org.kaazing.k3po.driver.internal.netty.channel.Channels;
 
 public class HttpChildChannelSource extends HttpChannelHandler {
 
@@ -125,9 +126,7 @@ public class HttpChildChannelSource extends HttpChannelHandler {
             case CONTENT_COMPLETE:
                 break;
             default:
-                ChannelException exception = new ChannelException("Channel closed unexpectedly");
-                exception.fillInStackTrace();
-                fireExceptionCaught(httpChildChannel, exception);
+                Channels.fireChannelAborted(httpChildChannel);
                 break;
             }
         }
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpClientChannelSink.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpClientChannelSink.java
@@ -151,8 +151,8 @@ public class HttpClientChannelSink extends AbstractChannelSink {
                     httpConnectChannel.setRemoteAddress(httpRemoteAddress);
                     httpConnectChannel.setConnected();
 
-                    fireChannelConnected(httpConnectChannel, httpRemoteAddress);
                     httpConnectFuture.setSuccess();
+                    fireChannelConnected(httpConnectChannel, httpRemoteAddress);
                 } else {
                     httpConnectFuture.setFailure(connectFuture.getCause());
                 }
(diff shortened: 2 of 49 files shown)


dpwspoon:
This PR very likely fixes: https://github.com/k3po/k3po/issues/297

dpwspoon:
I think this will also fix https://github.com/k3po/k3po/issues/202

mjolie:
Following this PR.

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
```diff
@@ -125,9 +126,7 @@ public void channelClosed(ChannelHandlerContext ctx, ChannelStateEvent e) throws
             case CONTENT_COMPLETE:
                 break;
             default:
-                ChannelException exception = new ChannelException("Channel closed unexpectedly");
-                exception.fillInStackTrace();
-                fireExceptionCaught(httpChildChannel, exception);
+                Channels.fireChannelAborted(httpChildChannel);
```
cmebarrow: Why is this only done in the write status switch statement and not the switch on the read status (which still throws the "Channel closed unexpectedly" exception} ?

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpClientChannelSink.java
```diff
@@ -261,6 +263,21 @@ protected void flushRequested(ChannelPipeline pipeline, FlushEvent evt) throws E
     }
 
     @Override
+    protected void abortRequested(ChannelPipeline pipeline, final AbortEvent evt) throws Exception {
+        HttpClientChannel channel = (HttpClientChannel) pipeline.getChannel();
+        ChannelFuture flushFuture = Channels.future(channel);
+        flushRequested(channel, flushFuture);
```
cmebarrow: Why would abort do a flush?

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
dpwspoon: Good catch.  I'll added it for completeness.  Note, I believe it is impossible for that switch to ever be hit right now with a valid rpt.  And it's important to understand the limitation of the abort as implemented.  Specifically the enum is `REQUEST, CONTENT_CHUNKED, CONTENT_COMPLETE, UPGRADED `.  It is not possible to abort a request half way through, as that is the design of k3po.  I.E. you can't script an abort in the request as follows, because the http request is just configuration data that is flushed all to the wire at once.  Also, I don't believe chunking is supported yet, see https://github.com/k3po/k3po/issues/136, https://github.com/k3po/k3po/issues/11, and https://github.com/k3po/k3po/issues/233.

```
connect http://localhost:8080/path
connected
write method "GET"
abort
write header "Some" "header"

```

Thus I don't know how I'll test it.  Maybe the following:

```
connect http://localhost:8080/path
connected
abort

```

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpClientChannelSink.java
dpwspoon: I think an implicit flush may be better here, as I can't see a case where a script author would want to add a write followed by an abort, and not actually have the write make it to the wire.  

Note: We make this same assumption when the connection is upgraded or in a connection close streaming state and the author has a `close`.

If you agree I'll add a comment here explaining why

driver/src/test/resources/org/kaazing/k3po/driver/internal/http/http.client.channel.abort/request.rpt
```diff
@@ -0,0 +1,24 @@
...
+#
+
+accept tcp://localhost:8080
+accepted
+connected
+read "GET /path HTTP/1.1\r\n"
+read "Host: localhost:8080\r\n"
+read "\r\n"
+
+closed
```
cmebarrow: why isn't it aborted?

driver/src/test/java/org/kaazing/k3po/driver/internal/http/HttpIT.java
```diff
@@ -255,4 +255,39 @@ public void testReadContentLengthViaAVariable() throws Exception {
         k3po.finish();
     }
 
+    @Test
+    @TestSpecification({
+        "http.server.channel.abort/request",
+        "http.server.channel.abort/response"
+    })
+    public void testHttpServerChannelAbort() throws Exception {
```
cmebarrow: No need to have test method names including the word "test"

driver/src/test/resources/org/kaazing/k3po/driver/internal/http/http.client.channel.abort/request.rpt
```diff
@@ -0,0 +1,24 @@
+#
```
cmebarrow: This is the response script isn't it?

driver/src/test/resources/org/kaazing/k3po/driver/internal/http/http.client.channel.aborted/request.rpt
```diff
@@ -0,0 +1,24 @@
+#
```
cmebarrow: This is the response script (need to flip the names)

driver/src/test/resources/org/kaazing/k3po/driver/internal/http/http.client.channel.abort/request.rpt
dpwspoon: Abort is supposed to kill the transport beneath it.  So for tcp we have no API on that.  I didn't implement anything for it, and as of now if someone used it in a script it would have no effect.  I also did not implement anything for Agrona (File transport).  Not sure what abort would mean in that case either.

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpChildChannelSource.java
cmebarrow: Yes that seems reasonable

driver/src/main/java/org/kaazing/k3po/driver/internal/netty/bootstrap/http/HttpClientChannelSink.java
cmebarrow: I don't mind either way as long as we document the behavior clearly (in the wiki)

cmebarrow:
@dpwspoon Done reviewing. Let me know when you ready for me to look again and merge.

dpwspoon:
@cmebarrow applied feedback.  Thanks

dpwspoon:
Hey, I found a bug, where when the abort is sent on the server channel it closes the parent.  Thereby unbinding all server channels.  Working on a fix and tests now.

dpwspoon:
@cmebarroI I fixed the bug and this is now ready for review

dpwspoon:
@cmebarrow I fixed the bug and this is ready for review

PR #304, opened 2016-02-04
Reverted the port numbers and the paths for extended handshake script…

sanjay-saxena:
…s for conformance in lieu of reusability.

### DIFF
--- a/specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
+++ b/specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
@@ -15,7 +15,7 @@
 #
 
 property writeMask ${httpx:randomBytes(4)}
-connect http://localhost:8000/echo
+connect http://localhost:8080/path
 connected
 
 write method "GET"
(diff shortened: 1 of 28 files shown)


specification/httpx/src/main/scripts/org/kaazing/specification/httpx/opening/extended.handshake.protocol.negotiated/response.rpt
```diff
@@ -14,13 +14,13 @@
...
 
-accept http://localhost:8001/echo
+accept http://localhost:8000/echo
 accepted
 connected
 
 read method "GET"
 read version "HTTP/1.1"
-read header "Host" "localhost:8001"
+read header "Host" "localhost:800"
```
dpwspoon: missing 8000

specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
```diff
@@ -15,7 +15,7 @@
 #
 
 property writeMask ${httpx:randomBytes(4)}
-connect http://localhost:8001/echo
```
dpwspoon: It might be best to use 8080 and "path" like the other scripts.  Just a suggestion though

PR #309, opened 2016-02-25
Add ability to override property from control

dpwspoon:


dpwspoon:
@cmebarrow this is ready for review

control/src/test/java/org/kaazing/k3po/control/internal/ControlTest.java
```diff
@@ -127,6 +127,7 @@ public void shouldWritePrepareCommand() throws Exception {
         final byte[] expectedPrepare =
                 ("PREPARE\n" +
                  "version:2.0\n" +
+                 "content-length:0\n" +
```
cmebarrow: Is this a backward compatible change to the protocol? Maybe bump protocol version?

examples/src/test/java/org/kaazing/k3po/examples/internal/PropertyOverrideIT.java
```diff
@@ -0,0 +1,74 @@
...
+    private final K3poRule k3po = new K3poRule().scriptProperty("RESPONSE 'Let\\'s take a selfie'");
+
+    private final TestRule timeout = new DisableOnDebug(new Timeout(5, SECONDS));
+
+    @Rule
+    public final TestRule chain = RuleChain.outerRule(k3po).around(timeout);
+
+    @Test
+    @Specification("server.hello.world")
+    @ScriptProperty("location 'tcp://localhost:8005'")
```
cmebarrow: Would  the following also work:`"location \"tcp://localhost:8005\""` ?

examples/src/test/scripts/org/kaazing/k3po/examples/internal/server.hello.world.rpt
```diff
@@ -14,9 +14,11 @@
...
-accept tcp://localhost:8001
+property location "tcp://localhost:8001"
+property RESPONSE "hello client"
+accept ${location}
 accepted
 connected
 read "hello world"
-write "hello client"
+write ${RESPONSE}
 closed
```
cmebarrow: Missing EOL

junit/src/main/java/org/kaazing/k3po/junit/rules/K3poRule.java
```diff
@@ -65,12 +67,14 @@
...
     private URL controlURL;
     private SpecificationStatement statement;
+    private List<String> classOverriddenProperties;
 
     /**
      * Allocates a new K3poRule.
      */
     public K3poRule() {
         latch = new Latch();
+        classOverriddenProperties = new ArrayList<>();
```
cmebarrow: Why do we need this? Nothing ever seems to get set into it, and the annotation definition appears to imply it can only be used on a method. Having said that, maybe it would be useful to allow it at class level too. Was that the intention?

cmebarrow: OK missed it, sorry below, sorry. So are we supporting class level annotation?

cmebarrow: OK so I see there is a test case too, using the K3poRule.scriptProperty method. So ignore all these comments!

examples/src/test/scripts/org/kaazing/k3po/examples/internal/server.hello.world.rpt
```diff
@@ -14,9 +14,11 @@
 # limitations under the License.
 #
 
-accept tcp://localhost:8001
+property location "tcp://localhost:8001"
+property RESPONSE "hello client"
```
cmebarrow: How about including a third property that is not overridden? Just to make sure all the properties don't get replaced.

specification/k3po.control/src/main/scripts/org/kaazing/specification/control/accept.finished.rpt
```diff
@@ -25,10 +25,11 @@ read "name:org/kaazing/specification/control/assets/client\n"
 read "\n"
 
 write "PREPARED\n"
-write "content-length:174\n"
+write "content-length:218\n"
```
cmebarrow: Why we changing existing test scripts?

specification/k3po.control/src/main/scripts/org/kaazing/specification/control/accept.finished.with.override.properties.rpt
```diff
@@ -0,0 +1,75 @@
+#
+# Copyright 2007-2015, Kaazing Corporation. All rights reserved.
```
cmebarrow: This script seems to be without override properties?

specification/k3po.control/src/main/scripts/org/kaazing/specification/control/accept.finished.with.barriers.rpt
```diff
@@ -27,10 +27,11 @@ read "\n"
 write "PREPARED\n"
 write "barrier:AWAITING_BARRIER\n"
 write "barrier:NOTIFYING_BARRIER\n"
-write "content-length:233\n"
```
cmebarrow: Why do we need to alter this script?

control/src/test/java/org/kaazing/k3po/control/internal/ControlTest.java
dpwspoon: I don't think this is backwards incompatible, at least it should not be.  And this test should be testing that: https://github.com/k3po/k3po/blob/develop/specification/k3po.control/src/main/scripts/org/kaazing/specification/control/accept.finished.empty.rpt

examples/src/test/java/org/kaazing/k3po/examples/internal/PropertyOverrideIT.java
dpwspoon: yes

PR #311, opened 2016-03-01
Adding couple of negative tests for origin security

jitsni:


specification/httpxe/src/main/scripts/org/kaazing/specification/httpxe/origin/bad.request.with.origin.header.and.x.origin.header/request.rpt
```diff
@@ -0,0 +1,28 @@
...
+#
+
+connect http://localhost:8000/path
+connected
+
+write method "POST"
+write version "HTTP/1.1"
+write header host
+write header "Origin" "http://unknownsource.example.com:80"
+write header "X-Origin" "http://unknownsource.example.com:80"
```
cmebarrow: Is this a bad request because Origin (and X-Origin) do not match the host?

specification/httpxe/src/main/scripts/org/kaazing/specification/httpxe/origin/bad.request.with.origin.header/request.rpt
```diff
@@ -0,0 +1,27 @@
+#
```
cmebarrow: Similary consider renaming to unauthorized.request.with.origin.header

cmebarrow:
@jitsni Done reviewing, looks good, just consider renaming the two new test cases

PR #312, opened 2016-03-03
Adding some more negative tests for httpxe

jitsni:
Adding some more negative tests for:
* request method is path-encoded, Origin header not present, X-Origin present, X-Origin-http-xxx header not present or xxx part does not match X-Origin
* request method is path-encoded, Origin header not present, X-Origin present, X-Origin-http-xxx header xxx part does not match X-Origin
* X-Origin present and differs from the Origin header, and the request is not path-encoded, and the request is NOT determined to be a same-origin request by comparing value of the Origin header with the Host header
* X-Origin header is present and the Origin header is not present, and the request is not path-encoded, and the request is NOT determined to be a same-origin request by comparing value of the Referer header with the Host header

One thing different is that response status code is 403 (gateway does that too). It computes effective Origin header and runs through constraints.

It seems to me that we should separate out the functionality: method is path-encoded should mean to represent method. It shouldn't have any relation with cross origin security.

### DIFF
--- a/specification/httpxe/SPEC.md
+++ b/specification/httpxe/SPEC.md
@@ -571,7 +571,7 @@ POST /path HTTP/1.1
 Host: target.example.com:80
 Origin: http://source.example.com:80
 ```
-If the authoritative source origin cannot be determined, then the server MUST response with status code `400 Bad Request`.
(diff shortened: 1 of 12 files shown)


cmebarrow:
Need to update SPEC.md to reword the sentence "If the authoritative source origin cannot be determined, then the server MUST response with status code 400 Bad Request" with something allowing either 400 or 403.

PR #314, opened 2016-03-03
Added support for http chunking trailers, (and worked on adding extensions), 

dpwspoon:
Note, a few changes of mine that fixed some of the rfc 7230 (http) spec tests ended up getting bunched with this.  It was while I was wokring on that, that I noticed the need for this new feature.


dpwspoon:
@jitsni or @cmebarrow can one of you review when you get a chance.  Thanks

PR #315, opened 2016-03-07
http method testing for GET,POST,OUT,HEAD

ilyaanisimov-kaazing:
rfc 7231 tests

### DIFF
--- a/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/method.definitions/post/response.rpt
+++ b/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/method.definitions/post/response.rpt
@@ -24,6 +24,5 @@ read "content"
 read closed
 	
 write status "200" "OK"
-write header content-length
 write close
 
--- a/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/method.definitions/post/response.rpt
+++ b/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/method.definitions/post/response.rpt
@@ -1,0 +1,29 @@
+#
+# Copyright 2007-2015, Kaazing Corporation. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


specification/http/src/test/java/org/kaazing/specification/http/rfc7231/RequestMethodsIT.java
```diff
@@ -15,39 +15,78 @@
...
+        k3po.finish();
     }
 
     @Test
-    @Ignore("Not Implemented")
-    public void serverShouldRespondWith405ToUnrecognizedMethods() {
-        // 405 (Method Not Allowed)
+    @Specification({
+        "405/request",
+        "405/response" })
```
dpwspoon: We should rename this test to more properly reflect the RFC, i.e change Unrecognized to "NotAllowed" or something like that.

>When a request method is received that is known by an origin
   server but not allowed for the target resource, the origin server
   SHOULD respond with the 405 (Method Not Allowed) status code.

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/method.definitions/post/response.rpt
```diff
@@ -0,0 +1,29 @@
...
+accepted
+connected
+
+read method "POST"
+read header "Content-Length" "7"
+read "content"
+read closed
+	
+write status "200" "OK"
+write header content-length
```
dpwspoon: I don't think you need the `write header content-length` for most of these response scripts.  They don't hurt but they don't really add anything IMO

dpwspoon:
LGTM (Looks good to me)

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/method.definitions/post/response.rpt
ilyaanisimov-kaazing: Script fails if I don't have write header content-length. Let's discuss

ilyaanisimov-kaazing:
Ping @dpwspoon

PR #316, opened 2016-03-10
Added tests for WS compliance (maximum message size and maximum lifetime)

NicoletaOita:
Added tests for WS compliance (maximum message size and maximum lifetime)


specification/ws/src/main/scripts/org/kaazing/specification/ws/limits/should.fail.binary.payload.length.131073/handshake.request.and.frame.rpt
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+property key ${ws:handshakeKey()}
+property writeMask ${ws:randomBytes(4)}
+property client131073 ${ws:randomBytes(131073)}
+
+connect http://localhost:8081/echo
```
cmebarrow: Please make all your scripts use http://localhost:8080/echo to follow the convention in all of the other ws specification scripts. We can solve the problem of needing different Gateway configurations by using different test classes (see my comments on your other PR).

specification/ws/src/main/java/org/kaazing/specification/ws/internal/Functions.java
```diff
@@ -34,6 +34,29 @@
     private static final int MAX_ACCEPTABLE_HEADER_LENGTH = 200;
 
     @Function
+    public static String loginBase64Encoder(String login) {
+        byte[] bytes = login.getBytes();
+        return bytesToString(Base64.encode(bytes));
```
cmebarrow: You should be able to just do `return new String(Base64.encode(bytes));`, not need for the bytesToString function, which you can remove

specification/ws/src/main/java/org/kaazing/specification/ws/internal/Functions.java
```diff
@@ -34,6 +34,29 @@
     private static final int MAX_ACCEPTABLE_HEADER_LENGTH = 200;
 
     @Function
+    public static String loginBase64Encoder(String login) {
```
cmebarrow: This is a general method that converts a string into a base 64 encoded string, it's not specific to login, so I would recommend making in a more general to reflect that. How about encodeAsBase64 or base64Encode?

cmebarrow:
@NicoletaOita Please pull in the latest changes from the develop branch (for example by doing `git pull upstread develop` then `git push`. See also a few comments I added (just minor stuff).

specification/ws/src/main/scripts/org/kaazing/specification/ws/limits/should.fail.max.lifetime.exceeded/handshake.response.and.frame.rpt
```diff
@@ -0,0 +1,31 @@
...
+read header "Sec-WebSocket-Version" "13"
+read header "Authorization" ${authHeader}
+
+write status "101" "Switching Protocols"
+write version "HTTP/1.1"
+write header "Upgrade" "websocket"
+write header "Connection" "Upgrade"
+write header "Sec-WebSocket-Accept" ${ws:handshakeHash(key)}
+
+write notify HANDSHAKE_COMPLETE
```
cmebarrow: Don't notify the same barrier as the other script, that's confusing and could cause problems. Why do you need to notify a barrier here? If you really do need it, change the name to something like HANDSHAKE_RESPONSE_WRITTEN.

specification/ws/src/main/scripts/org/kaazing/specification/ws/limits/should.fail.max.lifetime.exceeded/handshake.response.and.frame.rpt
```diff
@@ -0,0 +1,31 @@
...
+
+# clean close
+write [0x88 0x02 0x03 0xe8]
+
+write notify CLOSE_RECEIVED
```
cmebarrow: Again you should not notify the same barrier here. I think you can just remove this write notify instruction.

specification/ws/src/main/scripts/org/kaazing/specification/ws/limits/should.fail.max.lifetime.exceeded/handshake.request.and.frame.rpt
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+property location 'http://localhost:8080/echoAuth'
+property key ${ws:handshakeKey()}
+property validCredentials ${ws:base64Encode("joe:welcome")}
+property authHeader ${ws:append("Basic ", validCredentials)}
```
cmebarrow: Nice, I really like the way you did this, much more readable than just putting in the hex base64 encoding value!

PR #317, opened 2016-03-10
Replaced status code 0xea (1002) with 0xf1 (1009) for tests with payloads larger than expected.

Anisotrop:
For compliance with https://tools.ietf.org/html/rfc6455#section-5.5 the payload of control frames must not exceed 125 bytes. The tests with 126 bytes length must return the response code 0xf1 (1009) as defined in https://tools.ietf.org/html/rfc6455#section-7.4.1.


jitsni:
I think 1002/Protocol Error is appropriate here. 1009 can be returned where payload is allowed but limited by some config. See autobahn case 2.5
http://autobahn.ws/testsuite/reports_20131013/clients/index.html#case_desc_2_5

Most browsers don't even send a status code.

cmebarrow:
@Anisotrop I agree with Jitu. 1009 is only appropriate for data frames which exceed a configured maximum length (as in the tests @NicoletaOita is adding), not control frames.

PR #320, opened 2016-03-22
Fixed the ws specification test scripts for max websocket message size exceeded

cmebarrow:
to expect the server to immediately abort the tcp connection rather than attempting to send a close frame, to conform more closely to RFC 6455 (see section "7.1.7.  Fail the WebSocket Connection" which mandates that the server stop reading incoming data when failing the websocket connection).

PLEASE DO NOT MERGE I am trying another solution.


cmebarrow:
Closing because this change is not needed. Server can just stop reading input and still send the close frame.

PR #322, opened 2016-03-30
Sec-WebSocket-Protocol header contains comma separated value list.

jitsni:
So separating them.

### DIFF
--- a/specification/ws/src/main/scripts/org/kaazing/specification/ws/opening/request.header.sec.websocket.protocol/handshake.response.rpt
+++ b/specification/ws/src/main/scripts/org/kaazing/specification/ws/opening/request.header.sec.websocket.protocol/handshake.response.rpt
@@ -26,7 +26,8 @@ read header "Upgrade" /(?i:websocket)/
 read header "Connection" /(?i:Upgrade)/
 read header "Sec-WebSocket-Key" /(?<key>[a-zA-Z0-9+\/=]{24})/
 read header "Sec-WebSocket-Version" "13"
-read header "Sec-WebSocket-Protocol" "primary, secondary"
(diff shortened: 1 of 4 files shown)


specification/ws/src/main/scripts/org/kaazing/specification/ws/opening/request.header.sec.websocket.protocol/handshake.response.rpt
```diff
@@ -26,7 +26,8 @@ read header "Upgrade" /(?i:websocket)/
 read header "Connection" /(?i:Upgrade)/
 read header "Sec-WebSocket-Key" /(?<key>[a-zA-Z0-9+\/=]{24})/
 read header "Sec-WebSocket-Version" "13"
-read header "Sec-WebSocket-Protocol" "primary, secondary"
+read header "Sec-WebSocket-Protocol" "primary"
+read header "Sec-WebSocket-Protocol" "secondary"
```
dpwspoon: Why is this needed?  If I recall correctly the comma separated http headers are equivalent to this, logically from http rfc

jitsni: Does k3po keep track of which headers are comma separated headers ? Not all headers are comma separated ones, for e.g even though Date contains values with comma, it is not a comma separated one. Without that knowledge it cannot distinguish between

```
read header "header" "value1"
read header "header" "value2"
```
and

```
read header "header" "value1, value2"
```

dpwspoon: got it, yeah I had forgotten about that

PR #325, opened 2016-04-06
Caught exceptions that cause catostrophic failures in K3po and thus c…

dpwspoon:
…ause junit to error out.  When in fact the Junit Assume or ignore can cause these in the first place.  It would be better to revisit all these places and ensure they can always clean up properly

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/Robot.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/Robot.java
@@ -358,8 +358,12 @@ public class Robot {
                 server.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
             }
             for (ClientBootstrapResolver clientResolver : configuration.getClientResolvers()) {
-                ClientBootstrap client = clientResolver.resolve();
-                client.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
+                try {
+                    ClientBootstrap client = clientResolver.resolve();
+                    client.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
+                } catch (RuntimeException e) {
+                    LOGGER.warn("Exception caught while trying to stop client pipelies: " + e);
+                }
             }
 
             // remove each handler from the configuration pipelines
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/Robot.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/Robot.java
@@ -354,15 +354,19 @@ public class Robot {
 
             // clear out the pipelines for new connections to avoid impacting the observed script
             for (ServerBootstrapResolver serverResolver : configuration.getServerResolvers()) {
-                ServerBootstrap server = serverResolver.resolve();
-                server.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
+                try {
+                    ServerBootstrap server = serverResolver.resolve();
+                    server.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
+                } catch (RuntimeException e) {
+                    LOGGER.warn("Exception caught while trying to stop server pipelies", e);
+                }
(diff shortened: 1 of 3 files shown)


dpwspoon:
@sbadugu Can we file a follow up task or story to add a test for junit Assume to k3po.  Apologies for not getting it done in this PR

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
```diff
@@ -31,6 +31,7 @@
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.logging.Logger;
```
jitsni: Not needed ?

driver/src/main/java/org/kaazing/k3po/driver/internal/resolver/LocationResolver.java
```diff
@@ -66,9 +70,14 @@ public URI visit(AstLocationExpression value, ELContext environment) throws Exce
...
             // TODO: Remove when JUEL sync bug is fixed https://github.com/k3po/k3po/issues/147
-            synchronized (environment) {
-                ValueExpression expression = value.getValue();
-                location = expression.getValue(environment);
+            try {
+                synchronized (environment) {
+                    ValueExpression expression = value.getValue();
+                    location = expression.getValue(environment);
+                }
+            } catch (javax.el.PropertyNotFoundException e) {
```
jitsni: can you add an import for PropertyNotFoundException ?

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
dpwspoon: fixed

driver/src/main/java/org/kaazing/k3po/driver/internal/resolver/LocationResolver.java
```diff
@@ -66,9 +70,14 @@ public URI visit(AstLocationExpression value, ELContext environment) throws Exce
...
+                LOGGER.warn(e.getMessage());
```
jitsni: Do you want to pass e to the logger method ?

driver/src/main/java/org/kaazing/k3po/driver/internal/resolver/LocationResolver.java
dpwspoon: fixed

driver/src/main/java/org/kaazing/k3po/driver/internal/Robot.java
```diff
@@ -358,8 +358,12 @@ private void stopConfiguration() throws Exception {
                 server.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
             }
             for (ClientBootstrapResolver clientResolver : configuration.getClientResolvers()) {
-                ClientBootstrap client = clientResolver.resolve();
-                client.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
+                try {
+                    ClientBootstrap client = clientResolver.resolve();
+                    client.setPipelineFactory(pipelineFactory(pipeline(closeOnExceptionHandler)));
+                } catch (RuntimeException e) {
```
jitsni: why does this throw an exception ? Also, do you require the same above for server resolvers ?

dpwspoon: > why does this throw an exception ?
Because the clientResolver resolves properties ${connectURI} that might not be set.  This is essentially one last kick for K3po to run the connects when they aren't explicitly called so we can return a correct diff.  IF the properties in the handlers are not resolved this throws a hard exception and shuts down the control channel, which isn't good. I'd be ok with adding a try catch to the serverResolve.resolve as well...

dpwspoon: I've added the catch to the server

Issue #200, opened 2015-06-22
read /:.*/ "[\n" sometimes not working correctly


cmebarrow:
This issue was found when working on [gateway issue #203](https://github.com/kaazing/gateway/issues/203). 
````
read /:.*/ "[\n"
````
when operating on an incoming packet containing ':' followed by 180 space characters followed by two newline characters ("\n\n" is sometimes reporting a test failure as follows:
````
 org.junit.ComparisonFailure: Specified behavior did not match expected:<...ad "\n"
read /:.*/ "[\n"
read "\n"
read "data:Kobe is greater than Griffen\n"
read "\n]"
> but was:<...ad "\n"
read /:.*/ "[ ]"
>
org.junit.ComparisonFailure: Specified behavior did not match expected:<...ad "\n"
read /:.*/ "[\n"
read "\n"
read "data:Kobe is greater than Griffen\n"
read "\n]"
> but was:<...ad "\n"
read /:.*/ "[ ]"
>
````

**Comment by cmebarrow** [comment]
I can reproduce this using a spec test I added to specification/sse. This is in the attached patch file (apply on top of k3po/k3po develop branch commit ) and in branch issue_200_test_case in my forked repository 
[k3po_issue_200_test_case.diff | uploaded via ZenHub](https://files.zenhub.io/5588f910c4315ca3337657a1)

**Comment by cmebarrow** [comment]
The following k3po class appears to handle regular expression matching on read:
https://github.com/k3po/k3po/blob/develop/driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/handler/codec/ReadRegexDecoder.java

**Comment by cmebarrow** [comment]
The bug is happening due to fragmentation. If only a portion of ":                 \n" is read,   e.g. ":   ", ReadRegexDecoder thinks the pattern is fully matched, so that decoder gets removed from the list of decoders. When the next package comes in ("         \n") the FixedByteDecoder for "\n" then fails because the first space character does not match "\n".

**Comment by cmebarrow** [comment]
Work around (or perhaps just the correct way to write the test script?) is to put 
```
read /.*\n/
```
instead of
```
read /.*/ "\n"
```

**Comment by cmebarrow** [comment]
It is not desirable that robot test scripts should be vulnerable to fragmentation. We should therefore fix k3po so syntax like
```
read /:.*/ "\n"
```
works the same way irrespective of whether the portion of data before the newline character is fragmented or not. We are therefore treating this issue as a bug.

[...]

PR #328, opened 2016-04-12
Removed all instance where fragmentation in regex could cause a test …

dpwspoon:
…failure

### DIFF
--- a/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/proxy.must.not.retry.non.idempotent.requests/proxy.rpt
+++ b/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/proxy.must.not.retry.non.idempotent.requests/proxy.rpt
@@ -18,8 +18,8 @@ accept tcp://localhost:8080
 accepted
 connected
 
-read "POST" /.*/ "\r\n"
-read "Host:" /.*/ "\r\n"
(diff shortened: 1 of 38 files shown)


sanjay-saxena:
Failed on Travis in BBOSH.

cmebarrow:
Should we just fix [read /:.*/ "\n" sometimes not working correctly #200](https://github.com/k3po/k3po/issues/200) instead?

dpwspoon:
@cmebarrow Why not both ? :-)

In a few cases I've seen people write scripts like the following...

```
read /regex/
read "\n"
```

So there will still be an educational issue here, though it will be less pronounced if we do proposed fix in https://github.com/k3po/k3po/issues/200.  And then they will always have this fallback if thats what they actually prefer

jitsni:
Regex matching is tricky if you don't know the end of input. Say the script
is
```
read /a*/
read a
```
If the data comes in different events, then you need to sufficiently wait
until all the data arrives. Then there are cases like below which are
ambiguous to match:

```
read /a*/
read a
```
Difficult to match
```
read /a*/
read /b*/
```

In some cases we may need to build a bigger regex to match stream of data.


On Tue, Apr 12, 2016 at 5:16 PM, David Witherspoon <notifications@github.com

dpwspoon:
For now I think we should merge this and always try to use the single regex with terminator included in the regex.  This PR makes that change across the board.

Issue #330, opened 2016-04-15
Sporadic NPE in Control.java on Ubuntu 14.04


sanjay-saxena:
NPE is observed when trying to build Gateway's `release/5.1.0` branch on Ubuntu 14.04 using JDK 8. Here is the stack trace:

```
shouldEstablishConnectionWithRandomCaseRequestHeaders(org.kaazing.gateway.transport.wsn.specification.ws.connector.OpeningHandshakeIT) starting
OpeningHandshakeIT.shouldEstablishConnectionWithRandomCaseRequestHeaders FAILED with exception java.lang.NullPointerException (0.00 secs)
java.lang.NullPointerException
	at org.kaazing.k3po.junit.shaded.control.internal.Control.writeCommand(Control.java:282)
	at org.kaazing.k3po.junit.shaded.control.internal.Control.dispose(Control.java:479)
	at org.kaazing.k3po.junit.rules.ScriptRunner.dispose(ScriptRunner.java:313)
	at org.kaazing.k3po.junit.rules.SpecificationStatement.evaluate(SpecificationStatement.java:126)
	at org.junit.rules.Verifier$1.evaluate(Verifier.java:35)
	at org.kaazing.gateway.transport.wsn.specification.ws.connector.WsnConnectorRule$ConnectorStatement.evaluate(WsnConnectorRule.java:136)
```

dpwspoon:
This exception shows that the control channel was terminated on k3po prior to it being cleaned properly.  We can guard against it in the k3po junit code, but its an indication of a larger problem on the k3po driver side

dpwspoon:
`OutputStream bytesOut = connection.getOutputStream();` connection is null

PR #331, opened 2016-04-15
Moved controller.dispose into proper try catch block

dpwspoon:


dpwspoon:
resolves https://github.com/k3po/k3po/issues/330

PR #333, opened 2016-04-18
Fix where barrierFuture should be listened to even if handler future …

dpwspoon:
…is in incomplete state, which can occur in some race conditions

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/handler/barrier/AwaitBarrierUpstreamHandler.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/handler/barrier/AwaitBarrierUpstreamHandler.java
@@ -50,10 +50,7 @@ public class AwaitBarrierUpstreamHandler extends AbstractBarrierHandler implemen
             public void operationComplete(final ChannelFuture f) throws Exception {
                 if (f.isCancelled()) {
                     handlerFuture.cancel();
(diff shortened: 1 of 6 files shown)


driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/handler/barrier/AwaitBarrierUpstreamHandler.java
```diff
@@ -48,15 +48,12 @@ public void prepareRequested(final ChannelHandlerContext ctx, PreparationEvent e
...
-                // If the pipeline was not complete successfully we dont want nor need to wait for the barrier
-                if (!f.isSuccess()) {
-                    if (f.isCancelled()) {
-                        handlerFuture.cancel();
-                    } else {
-                        handlerFuture.setFailure(f.getCause());
-                    }
+                if (f.isCancelled()) {
+                    handlerFuture.cancel();
+                } else if (!f.isDone() && !f.isSuccess()) {
```
jitsni: I am looking at ChannelFuture's javadoc. Since this code is in operationComplete(), it means that channel i/o operation is *completed*. I would expect that isDone() is true in this method. Also, something like following would be more readable (no ! in conditions)

```
if (f.isCancelled()) {
..
} else if (f.isSuccess()) {
...
} else {
  assert f.cause() != null;
...
}
```

jitsni:
Is the travis failure related to this change set ?

PR #335, opened 2016-04-19
Reverted fix as it was incorrect, events that arrive too early should fail

dpwspoon:


driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/handler/event/AbstractEventHandler.java
```diff
@@ -99,14 +99,7 @@ protected void handleUpstream1(ChannelHandlerContext ctx, ChannelEvent evt) thro
...
             ChannelFuture pipelineFuture = getPipelineFuture();
-            if (!pipelineFuture.isDone()) {
-                pipelineFuture.addListener(new ChannelFutureListener() {
-                    @Override
-                    public void operationComplete(ChannelFuture future) throws Exception {
-                        handleUpstream1(ctx, evt);
-                    }
-                });
-            } else if (!pipelineFuture.isSuccess()) {
+            if (!pipelineFuture.isSuccess()) {
```
jitsni: Is this correct ? We don't know pipelineFuture is completed or not at this point. So pipelineFuture.isSuccess() is false doesn't mean a failure, right ?

dpwspoon: `// expected event arrived too early`

PR #336, opened 2016-04-19
Removed checking of event to early as this is done with implicit barr…

dpwspoon:
…iers


driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/handler/event/AbstractEventHandler.java
```diff
@@ -99,14 +99,7 @@ protected void handleUpstream1(ChannelHandlerContext ctx, ChannelEvent evt) thro
             handleUnexpectedEvent(ctx, evt);
         } else {
             ChannelFuture pipelineFuture = getPipelineFuture();
```
jitsni: Looks like you can remove pipelineFuture local variable.

PR #341, opened 2016-06-25
UDP support

jitsni:
specification.udp is work in progress. It will be completed in next PR


driver/pom.xml
```diff
@@ -109,10 +109,12 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
             </plugin>
+<!--
```
dpwspoon: We cna remove this?  If this is to disable it I suggest adding it to top level and use the checkstyle property to disable it

driver/src/test/java/org/kaazing/k3po/driver/internal/RobotIT.java
```diff
@@ -468,6 +471,129 @@ public void shouldEcho() throws Exception {
     }
 
     @Test
```
dpwspoon: I think we were trying to implement these kind of tests using the `@TestSpecification` framework that mimics the Junit API.  I guess it's ok to do it this way, but was there a specific reason you chose this class functionally over the other.  If not, lets add deprecated to this class and explain the reason (Test framework is similar to junit and takes nearly identical code paths).  I believe K3poIT has examples...

dpwspoon:
Very nice work @jitsni, minor feedback...

dpwspoon:
Also, have we considered Multicast, will this work for that use case?

jitsni:
@dpwspoon It doesn't work for multicast. @jfallows and I discussed a bit about that and it needs a different way compared to udp broadcast. But I didn't think completely about multicast though.

driver/src/test/java/org/kaazing/k3po/driver/internal/RobotIT.java
jitsni: I just saw that some tcp tests are there in RobotIT and I added them. If you want I can create UdpIT for these two tests using @TestSpecification.

dpwspoon: If you don't mind please do, these test rely on an protected method in Robot only there for them that I would like to delete at some point.

PR #347, opened 2016-07-22
Upgrade to Agrona 0.5.2, consuming Agrona package name change

jfallows:


jitsni:
Looks ok to me. I wanted to do the same for gateway as I am using OneToOneRingBuffer

PR #352, opened 2016-08-10
fix httpx k3po scripts masking

danibusu:
@dpwspoon While investigating ticket [434](https://github.com/kaazing/tickets/issues/434), I realized that the k3po scripts in scenario "client.sends.message.between.opening.and.extended.handshake" were wrong, because the client was sending unmasked frames, which the server should have been rejecting.
I modified the scripts for the masking issue, but I also changed the server response to match the current behavior of the gateway.
The httpx spec doesn't seem clear in this aspect, so I'm not sure which is the intended response. After detecting that the client has sent a different WS frame than the extended handshake request, should the server respond with a WS frame containing an encapsulated 400 Bad Request HTTP response? Or should it send a WS close frame?
Thanks!

### DIFF
--- a/specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
+++ b/specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
@@ -36,14 +36,23 @@ read header "Sec-WebSocket-Protocol" "x-kaazing-handshake"
 read header "Server" /.+/
 read header "Date" /.+/
 
-write [0x82 0x05] ${writeMask}
+write [0x82 0x85] ${writeMask}
 write option mask ${writeMask}
 write "Hello"
 write option mask [0x00 0x00 0x00 0x00]
 
-read [0x88 0x02 0x03 0xe8]
+# 400 Bad Request
+read [0x82 0x7E 0x00 0xAD]
+read "HTTP/1.1 400 Bad Request\r\n"
+read "Content-Type: text/html\r\n"
+read /Date:.+\r\n/
+read /Content-Length: \d+\r\n/
+read "\r\n"
+read "<html><head></head><body><h1>400 Bad Request</h1></body></html>"
 
-write [0x88 0x02] ${writeMask}
+write [0x88 0x82] ${writeMask}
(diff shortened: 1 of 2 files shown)


dpwspoon:
> I modified the scripts for the masking issue, but I also changed the server response to match the current behavior of the gateway.

Excellent find.  Please check that we have a script testing that the server rejects unmasked client frames in the specification wsn.

> The httpx spec doesn't seem clear in this aspect, so I'm not sure which is the intended response. After detecting that the client has sent a different WS frame than the extended handshake request, should the server respond with a WS frame containing an encapsulated 400 Bad Request HTTP response? Or should it send a WS close frame?

To me, I try and to think of Httpx as HTTP over WS, which is analogous to HTTP over TCP.  The websocket not being masked is a WS error and analogous to a unrecoverable TCP error.  As such HTTP can not use the now defunct WS channel to send an error code.  The 400 can not be sent.  Error frame with WS close is correct

specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
```diff
@@ -1,49 +1,58 @@
...
+read header "Sec-WebSocket-Protocol" "x-kaazing-handshake"
+read header "Server" /.+/
+read header "Date" /.+/
+
+write [0x82 0x85] ${writeMask}
+write option mask ${writeMask}
+write "Hello"
+write option mask [0x00 0x00 0x00 0x00]
+
+# 400 Bad Request
```
dpwspoon: should not have 400 bad request

dpwspoon:
Also, please CC @jfallows on the PR to fix the WSN unmasked frame PR

danibusu:
@dpwspoon @jfallows The script is already fixed in this pull request. The client is sending a masked WS frame, after the initial handshake, but before the extended handshake. The server is waiting for the GET request encapsulated in the WS frame, but instead receives a frame containing "Hello".
Seing how the x-kaazing-handshake sub-protocol was already negociated, is it ok for the server to respond with a 400 Bad Request to the unexpected frame? Or this behavior is expected after the extended handshake is completed?
Thanks!

dpwspoon:
Yes, this is fine.  I confused comment with the script in this PR

PR #357, opened 2016-08-29
Added function for IPv6 support in turn.proxy.

Anisotrop:
Should be able to test STUN support for IPv6.

### DIFF
--- a/specification/turn/src/main/java/org/kaazing/specification/turn/internal/Functions.java
+++ b/specification/turn/src/main/java/org/kaazing/specification/turn/internal/Functions.java
@@ -15,6 +15,9 @@
  */
 package org.kaazing.specification.turn.internal;
 
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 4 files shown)


Anisotrop:
@dpwspoon Hi David, I have created this PR to help QA check that IPv6 is supported by turn.proxy.

specification/turn/src/main/java/org/kaazing/specification/turn/internal/Functions.java
```diff
@@ -67,6 +70,34 @@
     }
 
     @Function
+    public static byte[] ipV6XOR(String ip, byte[] transactionId) {
+        InetAddress address;
+        try {
+            address = InetAddress.getByName(ip);
+        } catch (UnknownHostException e) {
+            e.printStackTrace(); // TODO replace with log if available
```
dpwspoon: If you throw the exception (maybe Runtime), then the k3po runtime should throw the exception and print the stack trace... thoughts?

specification/turn/src/main/java/org/kaazing/specification/turn/internal/Functions.java
```diff
@@ -67,6 +70,34 @@
...
+            return null;
+        }
+
+        if (!(address instanceof Inet6Address)) {
+            System.out.println("Address must be IPv6."); // TODO replace with log if available
```
dpwspoon: Same as above

dpwspoon:
Relevant part of [Spec](https://tools.ietf.org/html/rfc5389#section-15.2):

>
 X-Port is computed by taking the mapped port in host byte order,
   XOR'ing it with the most significant 16 bits of the magic cookie, and
   then the converting the result to network byte order.  If the IP
   address family is IPv4, X-Address is computed by taking the mapped IP
   address in host byte order, XOR'ing it with the magic cookie, and
   converting the result to network byte order.  If the IP address
   family is IPv6, X-Address is computed by taking the mapped IP address
   in host byte order, XOR'ing it with the concatenation of the magic
   cookie and the 96-bit transaction ID, and converting the result to
   network byte order.

Anisotrop:
@dpwspoon I have performed the changes, and execution can be traced to the function call when wrong parameters are given, but the error message does not seem to appear in the logs.

dpwspoon:
Thanks @Anisotrop, I've filed the ticket to track that bug here: https://github.com/k3po/k3po/issues/358.  Please review to see if the details I got were correct,  I added question to label for now as I'd like you to confirm the observed behavior.

PR #359, opened 2016-08-30
Turn tests for mapped address and masking key with ipv6.

NicoletaOita:
There are two tests, because mapped address works with some IPv6 addresses and doesn't work with others (the test that doesn't work for now is ignored).


dpwspoon:
@NicoletaOita: These tests are failing

NicoletaOita:
@dpwspoon I've ignored the MappedAddressWithIPv6IT test which was failing because of [issue #719] (https://github.com/kaazing/tickets/issues/719). I've ignored it in the gateway repo, but missed to ignore it in k3po. There was also a small mistake in one of the scripts which I've fixed. Please have a look again.

PR #362, opened 2016-09-12
Turn tests with channel data

NicoletaOita:


specification/turn/src/test/java/org/kaazing/specification/turn/PeerConnectionIT.java
```diff
@@ -115,5 +116,76 @@ public void shouldSuccessfullySendDataMessage() throws Exception {
...
+            "channel.data.message.without.create.permissions/response" })
+    public void shouldFailSendingDataMessageWithoutPermissions() throws Exception {
+        k3po.finish();
+    }
+
+    /**
+     * See <a href="https://tools.ietf.org/html/rfc5766">RFC 5766: Turn Protocol</a>.
+     */
+    @Test
+    @Ignore ("Please see issue #733: https://github.com/kaazing/tickets/issues/733")
```
dpwspoon: Why is this ignored here, shouldn't this be ignored in the project that picks up this test (i.e. the gateway?)

specification/turn/src/test/java/org/kaazing/specification/turn/PeerConnectionIT.java
```diff
@@ -115,5 +116,76 @@ public void shouldSuccessfullySendDataMessage() throws Exception {
...
+    public void shouldFailSendingDataMessageTooShortLength() throws Exception {
+        k3po.finish();
+    }
+
+    /**
+     * See <a href="https://tools.ietf.org/html/rfc5766">RFC 5766: Turn Protocol</a>.
+     */
+    @Test
+    @Ignore ("Please see issue #731: https://github.com/kaazing/tickets/issues/731")
+    @Specification({
```
dpwspoon: Same question

specification/turn/src/test/java/org/kaazing/specification/turn/PeerConnectionIT.java
```diff
@@ -115,5 +116,76 @@ public void shouldSuccessfullySendDataMessage() throws Exception {
...
+            "incorrect.channel.data.message.short.lenght/request",
```
dpwspoon: spelling length

specification/turn/src/test/java/org/kaazing/specification/turn/PeerConnectionIT.java
dpwspoon: I think I see.  Write tests like this as if you are testing a server.  I.e.  Bytes on the wire is what you code in scripts and then the reaction would not be in script.  i.e.
```
Bytes on the wire                                              Code API

--- don't read this request   --->   [ server ]  
--- read this request         --->                  ---> forwards second request in code 
```

Then when we test the proxy we do 

```
Bytes on the wire/k3po                                         Bytes on the wire / k3po

script with 
--- don't read this request   --->   [ server ]  
--- read this request         --->                  ---> script for normal request
```

Hope this makes sense, if not feel free to ping me

NicoletaOita:
@dpwspoon Thank you for the feedback, I've modified the scripts according to your explanations and also added the rest of the server tests that I've been working on. Please review again.

dpwspoon:
Added additional TURN protocol tests

PR #368, opened 2016-09-30
Moved test for the case when ChannelData messages are sent with a reserved channel number from k3po to gateway

mgherghe:


dpwspoon:
This is part of the Spec so they should be in K3po

PR #370, opened 2016-10-02
There is race between downstream and upstream connects :

jitsni:
There is race between downstream and upstream connects :

1. If upstream connects first and sends invalid request, the wseb session would be
closed. And the upstream connect would return 404
2. If the downstream connects first, the current script works
Now using barriers to fix the case 1


jitsni:
Exactly like https://github.com/k3po/k3po/pull/369

PR #372, opened 2016-10-11
Http.multi.auth

dpwspoon:


junit/src/main/java/org/kaazing/k3po/junit/annotation/ScriptProperty.java
```diff
@@ -23,13 +23,14 @@
 /**
  * {@code OverrideProperty} annotation overrides a property in the script.
  * Use format "key \"value\"", so property location "ws://localhost would be
- * @OverrideProperty( "location \"ws://localhost\"" )
+ * \@OverrideProperty( "location \"ws://localhost\"" )
```
jfallows: This should read `ScriptProperty` instead of `OverrideProperty`, right?
Also, if the example being used is a WebSocket URL, then we need a trailing `/` to make it valid.

junit/src/main/java/org/kaazing/k3po/junit/annotation/ScriptProperty.java
```diff
@@ -23,13 +23,14 @@
...
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.METHOD })
 public @interface ScriptProperty {
     /**
      * Gets the "key \"value\"" that the property is overridden with.
+     * @return array of "key \"value\"" to override in the script
      */
```
jfallows: Just curious, wouldn't it make sense just to let this `ScriptProperty` be specified once for each property with an explicit property name and property value? (rather than needing to parse the contents of each `String` to apply structure)

dpwspoon:
This built for me locally.  It looks like travis is having URL issues right now

PR #374, opened 2016-10-12
Turn.proxy.tcp.allocations

Anisotrop:
Add integration tests and script for no failures scenario.


dpwspoon:
This looks great!

dpwspoon:
We are probably missing tests in the SPEC for 

>
6.3.  New STUN Error Codes

>   446    Connection Already Exists
   447    Connection Timeout or Failure

PR #380, opened 2016-11-10
 case-insensitive websocket header names

jitsni:
Fixing the specification scripts for case-insensitive header names and values. This may fix https://api.travis-ci.org/jobs/174640231/log.txt?deansi=true

### DIFF
--- a/specification/ws/src/main/java/org/kaazing/specification/ws/internal/Functions.java
+++ b/specification/ws/src/main/java/org/kaazing/specification/ws/internal/Functions.java
@@ -15,6 +15,8 @@
  */
 package org.kaazing.specification.ws.internal;
 
+import static java.lang.Character.toLowerCase;
+import static java.lang.Character.toUpperCase;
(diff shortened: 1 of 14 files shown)


jfallows:
@jitsni: this is an excellent addition to our negative test coverage. 👍 

Kindly split and rename these to follow the naming pattern of the other script scenarios please:
* `request.header.upgrade.case.insensitive.websocket`
* `request.header.connection.case.insensitive.upgrade`
* `response.header.upgrade.case.insensitive.websocket`
* `response.header.connection.case.insensitive.upgrade`

...and then rename the function to `randomCaseNot` and ensure that the after randomizing the case of the supplied string, the result is not still equal to the supplied string, re-randomizing the case again if necessary.

Naturally, we would then use the function by passing in the canonical form, such as `${randomCaseNot('websocket')}` and `${randomCaseNot('Upgrade')}` to ensure that each test is distinct from the `connection.established` scenario.

jitsni:
@jfallows ready for one more round of review

PR #382, opened 2016-12-09
Adding a test case where udp writes are done together

jitsni:


specification/udp/src/main/scripts/org/kaazing/specification/udp/rfc768/concurrent.writes.together/server.rpt
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+#
+
+accept udp://localhost:8080
+       notify BOUND
+
+# Stream 1
+accepted
+connected
+read /(?<hello1>[a-zA-Z0-9]+)/
```
cmebarrow: I would recommend having the clients write text like "1Hello", "1Bye", with the variable part at the start, then you can make the RE expressions tighter, like
```
read /(?<hello1>[123]Hello)/
```
That will eliminate any possible issues like packets being coalesced, or k3po getting confused because of fragmentation, and also makes the script easy to understand.

cmebarrow:
LGTM, merging

PR #383, opened 2016-12-13
Control protocol specs

StCostea:


specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+## The Protocol
+
+The protocol is made up of Commands and Events.  Commands are actions that can be triggered by an external entity.
+Events are triggered by observed actions or state changes and are sent back to the external entity.  The protocol 
+is formatted by
+
+TODO normalize
+
+```
+Command / Event Name\n
```
dpwspoon: For normalization add something like

```
Control Message   =   command-line
                      *( header-field LF )
                      LF
                      [ message-body ]
```

```
command-line      = command LF
command              = token
```

```
header-field = field-name ":" field-value
```

`message-body`, `field-name`, `field-value`, and `token` are defined in [HTTP RFC](https://tools.ietf.org/html/rfc7230)

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+empty line
+content
+
+```
+
+headers are key value pairs separated by a ":" and delimited by a new line "\n"
+
+If the "content-length" header is specified, then optional content of the specified length can be added.
+
+Headers that are not understood / processed by a command can be ignored.
```
dpwspoon: I would probably suggest the same for content-length (if not understood than ignore)

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+
+
+### Commands
+
+##### PREPARE command
+
+###### Description
+This command will make the k3po server load and prepare a set of scripts for execution. The scripts are not yet executed, just loaded from
+disk and parsed. Additional override values for properties existing in the scripts can be given in the content of the command.
```
dpwspoon: Is this correct `Additional override values for properties existing in the scripts can be given in the content of the command.`  It might be, but if so perhaps add an example

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+
+###### Headers
+- version - the version of the protocol. Currently the only supported version is 2.0.
+- content-length - the length of the content of the command. This header is optional, should be present only when content is not empty.
+- origin - the root folder for the scripts to be executed. This header is optional
+- name - the name of the script to be executed. It is a relative path to the origin. Can appear more than once (several scripts to be executed).
+
+###### Content
+The content for prepare command is a set of properties with values that will overwrite the values in the script. The properties overwritten must
+exist in the scripts. Otherwise the prepare will fail.
```
dpwspoon: How does it fail?

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+
+
+##### START command
+
+###### Description
+This command will start the execution of the scripts that were sent in a previous PREPARE command. An error will occur if there is no PREPARE
```
dpwspoon: Start just does the clients (The `connects`),  perpare does the binds `accepts`

dpwspoon: Please use RFC language for clarity, ie.  An error will occur => An error MUST occur if there is no PREPARE, PREPARED

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+command send before the START command. If the START command was already sent, an error will occur. If the command is successful, the server
+sends back a STARTED event.
```
dpwspoon: Consider moving flow into general section, and then specify/enumerating commands.  Not sure if this is cleaner or not...

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+N/A
+
+###### Content
+N/A
+
+
+##### AWAIT command
+
+###### Description
+This command will respond with a NOTIFIED event after the barrier with the name sent in the header is triggered on the k3po server.
```
dpwspoon: Suggest standardizing terminology near start of doc: client and driver.  driver acts as a server.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
+# K3po Control Protocol
+
+## Abstract
+
+This document defines the K3po control protocol used to communicate with a remote K3po server.
```
dpwspoon: Please make it clear what the client and driver should do when commands are sent.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+- barrier - the name of the barrier that should be waited on
+
+###### Content
+N/A
+
+
+##### NOTIFIED command
+
+###### Description
+This command will trigger the barrier with the name sent in the header and will respond with a NOTIFIED event with the same barrier name.
```
dpwspoon: sent from client to driver?  What if barrier is not present?

dpwspoon: What if barrier is already triggered

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+This event is sent by the server when the server encountered an error while processing another command.
+
+###### Headers
+- content-length - the length of the content of the command. This header is optional, should be present only when content is not empty.
+- summary - a short description of the error (normally the error type)
+	
+###### Content
+The content for the event is the description of the exception in case we have an exception or a more detailed description in case there is no exception. 
+
+
```
dpwspoon: junit-lifecycle should not be present

specification/k3po.control/SPEC.md
StCostea: Any header will be ignored. This includes 'content-length'. I don't think it is ok to repeat the statement for a specific header.

specification/k3po.control/SPEC.md
StCostea: This is the usage for content in the prepare command (generated for annotation ScriptProperty). I will add an example.

specification/k3po.control/SPEC.md
StCostea: Returns an ERROR event instead of the PREPARED event.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+N/A
+
+###### Content
+N/A
+
+
+##### ABORT command
+
+###### Description
+This command will abort the current execution. It can be sent also after a PREPARE command (and before a START command).
```
StCostea: Corrected.

specification/k3po.control/SPEC.md
StCostea: There are many places were I used 'An error will occur'. Should I replace all of them ?

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,195 @@
...
+	
+###### Content
+The content for the event is the observed script after the execution on the k3po server. This script can be compared with the script in the
+prepared event to check if the test was successful. 
+
+
+##### ERROR event
+
+###### Description
+This event is sent by the server when the server encountered an error while processing another command.
```
StCostea: Currently, the client is sending a DISPOSE after an error event is received.

I think the server does nothing, I'm not sure. It remains in the same state as before, after sending the error event.

specification/k3po.control/SPEC.md
StCostea: You want me to replace server with driver in the entire document ?

specification/k3po.control/SPEC.md
StCostea: I'm not sure what do you mean by this. Client sends commands and the server/driver respond with events. This is already there, under 'The Protocol'.

specification/k3po.control/SPEC.md
StCostea: I just checked, the validation for barrier name is done on the client locally. It is not sent to the server.

If the barrier is triggered, the NOTIFIED event is sent immediately (it is a listener on the barrier on the server).

specification/k3po.control/SPEC.md
dpwspoon: I was thinking "FINISH" was a command but now I see it is "FINISHED".  I might be dense but the only spot where I see this is 

>The protocol is made up of Commands and Events.  Commands are actions that can be triggered by an external entity.
 Events are triggered by observed actions or state changes and are sent back to the external entity.  The protocol 
 is formatted by

where the external entity thing is confusing.  External could be the thing under tests, again I suggest standardizing the terminology.

specification/k3po.control/SPEC.md
dpwspoon: Sorry, to be clearer, we should specify what a command should do if there is content, when an implementation does not syntactically understand what that content means.  I suspect, (though haven't checked) that implementations would need a different if/catch for when they don't understand a header vs when they don't expect content length.

dpwspoon: Confirmed that our implementation does have separate catch and exception for content.  See https://github.com/k3po/k3po/blob/develop/driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlDecoder.java line 316.  

If we specify that unrecognized content should be ignored then we need to change this implementation.  Which is probably better for future compatibility

specification/k3po.control/SPEC.md
dpwspoon: perfect

specification/k3po.control/SPEC.md
dpwspoon: please specify in doc.  Thanks

specification/k3po.control/SPEC.md
dpwspoon: Yes, we should use standard terminology through out the doc.  I'm okay with client/server but I see in some places we clarify as "k3po server", or say external entity.  Furthermore, I think client/server can be confusing in k3po parlance as the k3po driver acts as a client or server to test code (hence the author of this statement clarified with k3po server).  Thus driver and test framework or something like that would be best.  Whatever you do choose, though, just state at the top of the doc which is which and then there is no more ambiguity or need to over clarify in each instance.

specification/k3po.control/SPEC.md
dpwspoon: Please add to spec, and I suggest we send an ERROR

specification/k3po.control/SPEC.md
dpwspoon: delete the file and then we squash the commits in check in

dpwspoon: or preferably you do a rebase if you want to keep intermediate history

specification/k3po.control/SPEC.md
StCostea: Message body is expected only for commands PREPARE, FINISHED and ERROR (see line 188 in the same file). In case you will add it to another type of command, it will be interpreted as a new command. For example:
**START

this is the body.**
will be considered as a START command followed by an incorrect command 'this' that will return an error.
The problem is that there is no way to signal that a command is finished (that it contains or not a body). So the decision is taken based on the command type (those 3 only) and on the presence of content-length header. For any other case, the body is considered an error. Should I specify this in the specification ?

specification/k3po.control/SPEC.md
StCostea: Done.

specification/k3po.control/SPEC.md
StCostea: You are right. I have modified the description.

StCostea: Done

specification/k3po.control/SPEC.md
StCostea: I will add this in the flow, we will have to implement it afterwards.

specification/k3po.control/SPEC.md
StCostea: External entity was the client in this case. I will try to put driver and test framework, we will see how it looks.

specification/k3po.control/SPEC.md
StCostea: Done.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+The communication between the test framework and the driver is a conversation - the test framework sends a command, the driver responds with an
+event sent asynchronously when the action is finished. The only exception is the FINISHED event - the START command will have two events as a 
+response - STARTED when the execution of the scripts is started and FINISHED when the execution of the scripts is finished.
+
+Throughout the flow description bellow, **an error** represents an ERROR event sent by the driver.
+
+- The test framework initiates the conversation with a **PREPARE** command. This command will make the k3po server load and prepare a set
+of scripts for execution. The scripts are not yet executed, just loaded from disk and parsed. If there are ACCEPT instructions in the scripts,
+a bound on corresponding ports is performed. If the command is successful, the driver sends back a PREPARED event. In case of failure, an
```
dpwspoon: %s/bound/bind/ => replace bound with bind

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+ERROR event is sent.
+
+- If the PREPARED command is successful, a **START** command may be sent by the test framework. This command will start the execution of the
+ clients in the scripts that were sent in the PREPARE command. If the command is successful, the driver sends back a STARTED event. This 
+ command will also trigger the sending of a FINISHED event whenever the scripts finished execution on the k3po server.
+ 
+- If there is no PREPARE command send before the START command or if the PREPARE command was not successful an error with a message "Script has not
```
dpwspoon: %s/send/sent/

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+ been prepared or is still preparing" must occur. 
+ 
+- If the START command was already sent, an error with a message "Script has already been started" must occur.
+
+- Any time after the PREPARE command and before the DISPOSE command, the test framework can send an **AWAIT** command for a barrier existing
```
dpwspoon: I think it is important to note that The NOTIFIED event may never be sent back if that barrier is never triggered, so the test framework can't block on it

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+- Any time after the PREPARE command *and before DISPOSE?*, the test framework can send a **NOTIFY** command for a barrier existing in the scripts.
+ The command notifies the barrier and a response with a NOTIFIED event is received. In case the barrier does not exist, an error must be returned.
+
+- Any time after the PREPARE command *and before DISPOSE?*, the test framework can send an **ABORT** command. The driver must abort the current 
+execution. It may be sent any time after a PREPARE command.
+
+- Any time after the PREPARE command, the test framework can send a **DISPOSE** command. The driver must abort the execution and release the resources 
+in use. When done, a DISPOSED event must be sent back.
+
```
dpwspoon: maybe state that notify, notified, and abort recieved after a finished event is sent or received is silently discarded

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+TODO - after the test is finished, you can start a new test from the beginning, with a new PREPARE. This does not seem to work very well. Usually
```
dpwspoon: We should support this, but the current implementation requires closing the connection and opening a new one.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+- Any time after the PREPARE command, the test framework can send a **DISPOSE** command. The driver must abort the execution and release the resources 
+in use. When done, a DISPOSED event must be sent back.
+
+
+TODO - after the test is finished, you can start a new test from the beginning, with a new PREPARE. This does not seem to work very well. Usually 
+the server gets stuck. Sometimes even a new connection will not help, still the test will not work. Should we support this or should we force the 
+closing of the connection and opening a new one ? (refuse any command after a DISPOSE)
+
+
+TODO - at this moment, nothing is returned by the driver after an ABORT command. I think we should return something. A FINISHED is not acceptable,
```
dpwspoon: What benefit or edge cases do you think this would bring to the lifecycle of the driver or the test framework? 

From the test frameworks point of view, an abort is typically sent if the test times out.  This indicates that the test is not making progress on the executed script.  The driver receives the abort, and then sends a finish back with the progress that was made.   This way the test framework can say something like test failed due to script mismatched exception (shows diff), caused by test timeout.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+- If the START command was already sent, an error with a message "Script has already been started" must occur.
+
+- Any time after the PREPARE command and before the DISPOSE command, the test framework can send an **AWAIT** command for a barrier existing 
+in the scripts. The command receives a response with a NOTIFIED event only when the corresponding barrier is notified. In case the barrier 
+does not exist, an error must be returned.
+
+- Any time after the PREPARE command *and before DISPOSE?*, the test framework can send a **NOTIFY** command for a barrier existing in the scripts.
+ The command notifies the barrier and a response with a NOTIFIED event is received. In case the barrier does not exist, an error must be returned.
+
+- Any time after the PREPARE command *and before DISPOSE?*, the test framework can send an **ABORT** command. The driver must abort the current
```
dpwspoon: On receiving an ABORT The driver should return a finished command to indicate the script progress.  The client should then clean up with the normal DISPOSE DISPOSED closing handshake

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+- barrier - the name of the barrier that should be waited on
+
+###### Message body
+N/A
+
+
+##### NOTIFIED command
+
+###### Description
+This command will trigger on the k3po server the barrier with the name sent in the header and will have a response with a NOTIFIED event
```
dpwspoon: %s/server/driver

dpwspoon:
I don't think the .pngs are needed, but please either include them in the MD via a link or remove them

specification/k3po.control/SPEC.md
StCostea: This is not happening today. And what do you mean by discarded ? There should always be a response for a command, especially await and notify. 

I think it should be a no operation on the server and the response should come back. 
The only problem I see is with an await on a barrier that was never notified. If the driver responds with a notified it is not ok, as the barrier was not notified. And a notification will never come after the execution finished. I'm thinking about an error for an await on a barrier that was not notified if the execution is finished..

specification/k3po.control/SPEC.md
StCostea: Even with closing a connection and opening a new one sometimes there are problems. I will try to reproduce this with specific steps and investigate it.

If we are to support multiple tests on the same connection, we need to add clear specification on how the driver will work/respond for commands after a DISPOSE and before a new PREPARE.

specification/k3po.control/SPEC.md
StCostea: Ok, now I understand the role of abort command. I will update the spec accordingly.

Anyway, it seems that currently the driver does not send back the FINISHED event. I think I have seen it only once in many, many tests. We will have to correct that.

specification/k3po.control/SPEC.md
StCostea: Ignore this idea, now I understand the role of the ABORT command.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+N/A
+
+###### Message body
+N/A
+
+
+##### AWAIT command
+
+###### Description
+This command will have as a response a NOTIFIED event after the barrier with the name sent in the header is triggered on the k3po server.
```
StCostea: Changed.

specification/k3po.control/SPEC.md
StCostea: Done

specification/k3po.control/SPEC.md
StCostea: Done

specification/k3po.control/SPEC.md
StCostea: I just understood the way responses are given to ABORT - the FINISHED event is sent only once per test. In case the script is finished (and FINISHED sent), any ABORT command will have no response. If the script is not finished, the first ABORT will trigger a FINISHED event, subsequent ABORT commands will have no response. I have updated the flow accordingly.

specification/k3po.control/SPEC.md
```diff
@@ -0,0 +1,244 @@
...
+
+- Any time after the PREPARE command *and before DISPOSE?*, the test framework can send an **ABORT** command. The driver must abort the current 
+execution. If the execution of the scripts is already finished, no event is sent back. If the execution is aborted by this command, a FINISHED 
+event with the script that was executed so far must be returned.
+
+- Any time after the PREPARE command, the test framework can send a **DISPOSE** command. The driver must abort the execution and release the 
+resources in use. When done, a DISPOSED event must be sent back.
+
+
+TODO - Decision time :) - will we support multiple tests on the same connection or will we force closing the connection and reopening a new one ?
```
dpwspoon: I recommend supporting via SPEC and implementing in driver.  I don't think we need to implement this in JUNIT.  It's actually much easier to debug in wireshark when connection for control protocol is closely associated with the tests streams it is running

dpwspoon:
Please add a comment about what the default port is.  (I always forget and it would be good to have it documented cleanly.

dpwspoon:
@jfallows I think this is ready for your review

specification/k3po.control/SPEC.md
StCostea: Ok, I will update the specs.

PR #384, opened 2016-12-14
update message to work around HTTP method validation

danibusu:
Update the sent message to start with a letter other than ones of valid
HTTP methods to bypass gateway validation. In a real scenario, the
request would time-out.

### DIFF
--- a/specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
+++ b/specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
@@ -38,21 +38,12 @@ read header "Date" /.+/
 
 write [0x82 0x85] ${writeMask}
 write option mask ${writeMask}
-write "Hello"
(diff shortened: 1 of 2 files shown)


danibusu:
@jitsni This is the PR for the HTTP method validation issue.

sbadugu:
@jitsni Can you take an action item on this PR ?

specification/httpx/src/main/scripts/org/kaazing/specification/httpx/extended/client.sends.message.between.opening.and.extended.handshake/request.rpt
```diff
@@ -38,21 +38,12 @@ read header "Date" /.+/
 
 write [0x82 0x85] ${writeMask}
 write option mask ${writeMask}
-write "Hello"
+write "World"
```
jitsni: I see what happens in HTTP decoding when Hello is sent as method. Can you not just send "Hello World"

cmebarrow:
@danibusu Does this still need to get merged. If so, please answer Jitu's review comment.

dpwspoon:
not needed

PR #386, opened 2016-12-15
Added spec tests for redirection on http

dpwspoon:


specification/http/src/main/scripts/org/kaazing/specification/http/rfc7231/redirection/relative.location/request.rpt
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+connect http://localhost:8000/resource
+connected
+
+write method "GET"
```
jitsni: I think k3po is sends "GET http://localhost:8000/resource HTTP/1.1" on the wire (instead of "GET /resource HTTP/1.1"). This is not common. But adding "write Host" will generate "GET /resource HTTP/1.1"

dpwspoon:
@jitsni Thanks, I've made both improvements.  This is ready for another round of review

Issue #388, opened 2016-12-27
K3po driver is stuck if an abort is sent after an await on a barrier never triggered


StCostea:
If there is an await on a barrier that is never triggered, calling an abort will block the thread on waiting the barrier to be triggered. 
The following protocol commands will make the driver unresponsive:

```
PREPARE
version:2.0
name:org/kaazing/specification/control/assets/server
name:org/kaazing/specification/control/assets/client.with.barriers

AWAIT
barrier:NOTIFYING_BARRIER

ABORT

```
In the script `client.with.barriers`, the NOTIFYING_BARRIER is never notified.
The cause is the code in the listener `whenAbortedOrFinished `in class `ControlServerHandler`.

StCostea:
I am not sure what is the purpose of awaiting on all barriers before finishing. So I'm not sure how to fix it.

dpwspoon:
>  I am not sure what is the purpose of awaiting on all barriers before finishing. So I'm not sure how to fix it.

This sounds like an implementation bug to me, and probably it should not be implemented this way.

StCostea:
I have currently commented the await part in the listener. So we are just sending the FINISHED event. Is this ok as a solution ? 
With this code all specification tests are passing.

PR #389, opened 2017-01-04
K3po driver fixes according to new specs + some additional fixes

StCostea:
This includes the following:
- fix for https://github.com/k3po/k3po/issues/388 (removal of wait statements)
- modification of logic according to new k3po specs
- small fix to improve exception reporting - in case an InterruptedException is caused by a driver error, the driver exception should be reported instead.
- fix for https://github.com/k3po/k3po/issues/391 (interrupted status)
- fix for https://github.com/k3po/k3po/issues/393 (rewriting of ControlDecoder)

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
@@ -65,14 +65,21 @@ public class ControlServerHandler extends ControlUpstreamHandler {
     private static final Map<String, Object> EMPTY_ENVIRONMENT = Collections.<String, Object>emptyMap();
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(ControlServerHandler.class);
+    private static final String ERROR_MSG_NOT_PREPARED = "Script has not been prepared or is still preparing\n";
+    private static final String ERROR_MSG_ALREADY_PREPARED = "Script already prepared\n";
+    private static final String ERROR_MSG_ALREADY_STARTED = "Script has already been started\n";
 
     private Robot robot;
     private ChannelFutureListener whenAbortedOrFinished;
-    private BlockingQueue<CountDownLatch> notifiedLatches = new LinkedBlockingQueue<CountDownLatch>();
+    private BlockingQueue<CountDownLatch> notifiedLatches;
 
-    private final ChannelFuture channelClosedFuture = Channels.future(null);
+    private ChannelFuture channelClosedFuture;
 
     private ClassLoader scriptLoader;
+    
+    public ControlServerHandler() {
+        initialize();
+    }
 
     public void setScriptLoader(ClassLoader scriptLoader) {
         this.scriptLoader = scriptLoader;
@@ -92,6 +99,7 @@ public class ControlServerHandler extends ControlUpstreamHandler {
                 public void operationComplete(ChannelFuture future) throws Exception {
                     channelClosedFuture.setSuccess();
                     ctx.sendUpstream(e);
+                    initialize();
                 }
             });
         }
@@ -100,6 +108,10 @@ public class ControlServerHandler extends ControlUpstreamHandler {
 
     @Override
     public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) throws Exception {
+        if (robot != null && robot.getPreparedFuture() != null) {
+            sendErrorMessage(ctx, ERROR_MSG_ALREADY_PREPARED);
+            return;
+        }
 
         final PrepareMessage prepare = (PrepareMessage) evt.getMessage();
 
@@ -255,7 +267,16 @@ public class ControlServerHandler extends ControlUpstreamHandler {
 
     @Override
     public void startReceived(final ChannelHandlerContext ctx, MessageEvent evt) throws Exception {
-
+        if (robot == null || robot.getPreparedFuture() == null) {
+            sendErrorMessage(ctx, ERROR_MSG_NOT_PREPARED);
+            return;
+        }
+        
+        if (robot.getStartedFuture().isDone()) {
+            sendErrorMessage(ctx, ERROR_MSG_ALREADY_STARTED);
+            return;
+        }
+        
         try {
             ChannelFuture startFuture = robot.start();
             startFuture.addListener(new ChannelFutureListener() {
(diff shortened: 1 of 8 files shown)


junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -24,7 +24,7 @@
     private volatile State state;
     private volatile Exception exception;
 
-    private final CountDownLatch prepared;
+	private final CountDownLatch prepared;
```
dpwspoon: spaces, not tabs please

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -139,9 +139,10 @@ boolean hasException() {
 
     void notifyException(Exception exception) {
         this.exception = exception;
-        prepared.countDown();
-        startable.countDown();
-        finished.countDown();
+        // Commented because of issue https://github.com/k3po/k3po/issues/391
```
dpwspoon: I'm not following this, and based on the code I don't think you have gotten a fix yet for this, I would suggest pulling this out of the PR and putting in a different one

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -65,14 +65,21 @@
     private static final Map<String, Object> EMPTY_ENVIRONMENT = Collections.<String, Object>emptyMap();
```
dpwspoon: This looks good to me

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
```diff
@@ -55,10 +55,10 @@ public void evaluate() throws Throwable {
             // start the script execution
             new Thread(scriptFuture).start();
 
-            // wait for script to be prepared (all binds ready for incoming connections from statement)
-            latch.awaitPrepared();
-
             try {
+                // wait for script to be prepared (all binds ready for incoming connections from statement)
```
dpwspoon: This looks good to me

StCostea: About this one I'm not sure, it complicates the error handling. I'm trying an option to have a separate try-catch for awaitPrepared. When I'll get something working fine I'll commit it.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: The situation I encountered was the following: the prepare failed, the main thread was in awaitPrepared, the script thread calls notifyException. First it calls prepare.countDown() so the main thread gets unlocked and fails properly. But the script thread then calls testThread.interrupt() and the interrupt status is set for the thread. Because the main thread is not anymore in an await, nobody is clearing the interrupt status. 
The test ends with the expected result (failed in prepare), next test starts. When the first wait is encountered, the InterruptedException is thrown, although nothing caused it on the new test. So the new test fails with some strange exception (some InterruptException with no cause) and a NullPointerException is shown in the log because there is no connection. At least this is what happened in my tests. I guess that different exceptions/errors can occur.
So, in order to let the InterruptedException to occur, we need to skip unlocking the latches.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: Not sure how this got here. I'm using the Kaazing formatter. :(

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: What we need to check is what was the role of the commented lines. I don't see any.

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
```diff
@@ -112,10 +122,11 @@ public void evaluate() throws Throwable {
                 }
             }
 
+            // Stefan - not sure what the comment bellow is about. There is no join() method in the statement or rule class.
```
dpwspoon: Perhaps we can add tests here for wrapped rules on junit

dpwspoon:
Minor changes requested

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
StCostea: I think this is already there, in the assertTrue. What I don't understand is the join() method specified in the comments. There is no join() method in the statement or rule. Maybe it is an older comment and the join() method was removed ? Then we can remove or fix the comment.

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -393,9 +457,25 @@ private void sendErrorMessage(ChannelHandlerContext ctx, Throwable throwable) {
...
-            error.setSummary("Internal Error");
+            logger.error("Internal error. Sending error to client", throwable);
+            error.setSummary("Internal error");
             Channels.write(ctx, Channels.future(null), error);
         }
     }
+    
+    private void sendErrorMessage(ChannelHandlerContext ctx, String description) {
+        ErrorMessage error = new ErrorMessage();
+        error.setSummary("Internal error");
```
StCostea: Like in my answer [here](https://github.com/k3po/k3po/pull/390#discussion_r95751317) , I don't think this is a good idea. If you still think we should do this, let's discuss about it.

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
dpwspoon: > There is no join() method in the statement or rule. Maybe it is an older comment and the join() method was removed ?

I believe this is true.  But I'm not convinced whether it was correct to remove it or not.  

> I think this is already there, in the assertTrue.  

I'm not following where this is?  What I was trying to state was that the comment seems to indicate that the join is needed for cases where K3PO rule is in a [rule chain](http://www.hascode.com/2012/02/ordering-your-junit-rules-using-a-rulechain/).  Specifically, where rules after it (or maybe before depending on your mental ordering) complete before the K3po rule has completed execution.  If we have tests for this then we can safely remove that comment.  If we don't, perhaps we can add them and then safely remove the comment.

StCostea: The method `apply()` in the rules are called sequential, each one with the statement resulted from calling the previous rule. All this is in one thread, you cannot get the parameters to call the next rule before the previous one ends. So, I'm not sure what you mean by :

> rules after it (or maybe before depending on your mental ordering) complete before the K3po rule has completed execution

Even in `RuleChain`, the code in `apply()` is :
```
for (TestRule each : rulesStartingWithInnerMost) {
            base = each.apply(base, description);
        }
```
No multi-threading here...

I also think that the comment is about some checks that should be done in this piece of code (I suppose it is done in that assetTrue), not about some unit tests/integration tests.

StCostea: With some help from Costin, I have found an older commit (actually two) where a method called join was removed and replaced with start and finish. Take a look at them:
https://github.com/k3po/k3po/commit/eeb9d4127edcea5216f1c5cd94263f873c33b910
https://github.com/k3po/k3po/commit/e44f2e8b0b782faafb73edfdcd04d95d41f630e0

PR #390, opened 2017-01-05
[Merge after PR 383 and PR 389] Control protocol testcases

StCostea:
Must be merged (updated if required) after https://github.com/k3po/k3po/pull/383 and https://github.com/k3po/k3po/pull/389 are merged.

This PR contains the specification testcases plus test cases for the new version of driver.


specification/k3po.control/src/main/scripts/org/kaazing/specification/control/accept.error.if.no.prepare.rpt
```diff
@@ -0,0 +1,68 @@
...
+#
+
+accept tcp://localhost:12345
+accepted
+connected
+
+read "START\n"
+read "\n"
+
+write "ERROR\n"
```
dpwspoon: I would suggest that we change the SPEC, that an ERROR event will trigger the disconnect of the TCP connection and the client must start a new connection if it wishes to continue.  This allows us to tidy clean up, and not get into the error loops mentioned here

StCostea: This set of errors is in one test just for simplicity. Normally they will not occur like this. Anyway, the driver responds based on the state and I see no problem here.

If we are suppose to close connection on the driver, then we will have to stop the execution of the script. Also, if there is any result to be transmitted to the client it should not be done (a FINISHED is sent even if an ABORT occurs). 
From my point of view, this will complicate a lot the flow on the driver. Currently, it is the responsibility of the client to restart the connection if an error occurs, and the client should decide if he wants to restart or continue from this state. 

What you propose makes sense from a Junit test's point of view, because any kind of error will fail the test. But in case of a text connection to the driver, for example, or other type of usage, it might not be that useful. If I mistake the name of a barrier I need to restart from scratch instead of just trying again with the correct name.

dpwspoon:
Please fix travis build failures.

specification/k3po.control/src/main/scripts/org/kaazing/specification/control/accept.error.if.no.prepare.rpt
dpwspoon: IMHO, the driver is used to drive deterministic tests.  An error on the driver or by an invalid command etc always defines that the test is fatally flawed and can not recover, and the first thing to fix is what caused the error.  So I don't see a use case to support recovery when an error occurs.

Secondly, if there was such a use case.  I think it would be important to still have a way to identify a fatal error (don't retry) versus a recoverable error.  By defaulting everything to the fatal error case now, we can always open up/expand to the recoverable error case in the future.  But by defaulting to the recoverable error case now, means we need to add the mechanism for the fatal error case now, when it isn't clear we need to support such use cases yet.

StCostea:
We need first to merge https://github.com/k3po/k3po/issues/389 and then check if travis build still fails. Normally it should not fail, I checked it before doing the PR.

PR #394, opened 2017-01-17
Opensaml

DoruM:
Adds integration tests for SAMLLoginModule
Improves specification for SAML
Adds functions for encrypting and decrypting assertions


pom.xml
```diff
@@ -54,6 +54,7 @@
         <module>specification/ws.extensions/x-kaazing-idle-timeout</module>
         <module>specification/ws.extensions/x-kaazing-ping-pong</module>
         <module>specification/http.multi.auth</module>
+		<module>specification/saml.auth</module>
     </modules>
```
akrambek: Alignment issue.

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
```diff
@@ -0,0 +1,115 @@
...
+
+        String encoded_response = OpenSAMLUtils.compressAndEncodeString(OpenSAMLUtils.SAMLObjectToString(response), false);
+        return String.format("%s realm=\"%s\" saml-authn-request=\"%s\"", authnType, realm, encoded_response);
+    }
+
+    @Function
+    public static String generateResponse(String realm, String authnType) throws Exception {
+
+        GenerateResponse.intializeSAML();
+        Response response = GenerateResponse.generateResponse(false);
```
akrambek: There is typo in `intializeSAML ` should be `initialize`

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
```diff
@@ -0,0 +1,115 @@
...
+
+        String encoded_response = OpenSAMLUtils.compressAndEncodeString(OpenSAMLUtils.SAMLObjectToString(response), false);
+        return String.format("%s realm=\"%s\" saml-authn-request=\"%s\"", authnType, realm, encoded_response);
+    }
```
akrambek: I didn't get this why do you add `saml-authn-request` in the header I couldn't find in any spec.

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
DoruM: Huh? what do you mean? The function is initializeSAML

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
DoruM: Because this is how it works...We encode the challenge like this. THe saml-authn-request encoding contains the Base64 encoded Response. This is what we send to the Gateway

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
akrambek: `GenerateResponse.intializeSAML();` Do you see the same

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
akrambek: Yes my question is why do you add `saml-authn-request`? Sorry for the dump question but I couldn't find the reason so I am carious.

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
```diff
@@ -0,0 +1,115 @@
...
+        String encoded_request = OpenSAMLUtils.compressAndEncodeString(OpenSAMLUtils.SAMLObjectToString(request), false);
+
+        return String.format("%s realm=\"%s\" saml-authn-request=\"%s\"", authnType, realm, encoded_request);
+
+    }
+
+    /**
+     * This function checks if the string received as parameter encodes a AuthnRequest
+     * @param request - the challenge string received from the gateway
+     * @param compressed - indicates whether the challenge is encoded or not
```
akrambek: This `compressed ` param doesn't exist in the method signature.

specification/saml.auth/src/main/java/org/kaazing/specification/saml/internal/Functions.java
DoruM: We needed to find the actual Response. It is easiest to find it by searching for it using some regex or anything for that matter. It is more difficult if we don't have any type of anchor.

specification/saml.auth/src/main/scripts/org/kaazing/specification/http/saml/auth/request.with.secure.challenge.encrypted.response/server.rpt
```diff
@@ -0,0 +1,57 @@
...
+write status "101" "Switching Protocols"
+write version "HTTP/1.1"
+write header "Upgrade" "websocket"
+write header "Connection" "Upgrade"
+write header "Sec-WebSocket-Accept" ${ws:handshakeHash(key)}
+
+read [0x82 0x7d]
+read [0..125]
+
+write [0x88 0x02 0x03 0xea]
```
akrambek: Missing new line.

DoruM:
Moved to different repository

Issue #402, opened 2017-02-09
Unexpected event Notified


dpwspoon:
Timeout in test sending abort causes following exception log.  It should report script progress.

```
Running org.kaazing.specification.ws.extensions.x_kaazing_idle_timeout.IdleTimeoutExtensionIT
java.lang.IllegalArgumentException: Unexpected event kind: NOTIFIED
	at org.kaazing.k3po.junit.rules.ScriptRunner.dispose(ScriptRunner.java:227)
	at org.kaazing.k3po.junit.rules.SpecificationStatement.evaluate(SpecificationStatement.java:137)
	at org.junit.rules.Verifier$1.evaluate(Verifier.java:35)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:283)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:173)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:153)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:128)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:203)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:155)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
```

StCostea:
Added fixes:
[...]

PR #404, opened 2017-02-16
Finish execution before sending the DISPOSE command

StCostea:
Fix for issue https://github.com/k3po/k3po/issues/402.
Sometimes, when an abort was requested, there was an error after sending the DISPOSE command, because another event was in the incoming buffer in place of DISPOSED. 
Now we will wait for the execution to finish before sending the DISPOSE command.


dpwspoon:
Thanks @StCostea

Issue #391, opened 2017-01-05
Interrupted status true at the beginning of test


StCostea:
There were situations when an interrupt in a test that was failing was not processed and the next test was treating it as a received interrupt, failing with no clear cause.
This situation occurred sporadic as it was a race between the interrupt() call and the countDown() call on the prepared latch. Sometimes the wait exited with an InterruptedException, sometimes exited normally.

StCostea:
It seems that this issue related to interrupted flag has a bigger impact than expected. It is causing issues also in gateway tests where the flag is caught by the gateway rule. More investigations and an additional fix are required.

StCostea:
Added fixes:
- test thread is interrupted only from other threads (Latch class)
- clear the interrupted status in finally at the end of evaluate() method in the statement, to avoid it reaching other statements (like gateway)

PR #406, opened 2017-02-21
Additional fixes related to interrupted status and 'Unexpected event Notified'

StCostea:
- test thread is interrupted only from other threads (Latch class)
- the finished latch is opened in notifyFinished() even if in INIT state (Latch class)
- we keep reading events even after FINISHED is received, until a timeout occurs. This is done to clear the buffer before the dispose command (ScriptRunner class)
- clear the interrupted status in finally at the end of evaluate() method in the statement, to avoid it reaching other statements (like gateway)

Related to #402 and #391.


junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -93,7 +93,7 @@ void notifyFinished() {
         switch (state) {
         case INIT:
             notifyPrepared();
-            break;
+//            break;
```
dpwspoon: This state change to finished seems counter intuitive

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
```diff
@@ -137,12 +139,19 @@ public ScriptPair call() throws Exception {
...
+                        
+                        // we keep the result and read any other event is still in the buffer. We will return after the first timeout
+//                        return new ScriptPair(expectedScript, observedScript);
+                        finishedReceived = true;
+                        break;
                     default:
                         throw new IllegalArgumentException("Unrecognized event kind: " + event.getKind());
                     }
                 } catch (SocketTimeoutException e) {
+                    if (finishedReceived)
```
dpwspoon: does this require all cases to get a timeout exception?  Why not properly track the state changes instead.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: We have a call to notifyFinished, it seems normal to me that at the end of this call the state should be FINISHED.
Anyway, the real problem is that we need a flag to check if the thread reading the events in a loop (the ScriptRunner) is finished. I considered it to be the finished latch. Otherwise we will have to add another one, maybe something like stopped and set it when the notifyFinished is called. From my point of view, I don't see why we cannot use finished instead.

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
StCostea: The problem is that we can still receive events from the driver after the FINISHED events, no matter what is the state. In order to have the correct response DISPOSED when we send a DISPOSE, we need to clear the buffer of any other event still pending. This is why I delay returning the result until the timeout (buffer clear).

Other option is to have another loop in ScriptRunner.dispose() method to read until the DISPOSED event. This again will block the main thread waiting until a timeout occurs or the DISPOSED event is received.

Or just ignore and not read the DISPOSED event.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
dpwspoon: okay, perhaps renaming the latched would make it more clear to me.  Can we remove the commented out code (rather then comment it out)

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
dpwspoon: > The problem is that we can still receive events from the driver after the FINISHED events,  

Is that a spec violation by the driver?  Why not fix that?  Otherwise, we need this workaround on every client.

Waiting on the SocketTimeoutException seems to be a hack. Would greatly prefer another approach.  Waiting for confirmation on the `Disposed` is necessary as otherwise there is a race between tests on clean up of the driver and potential attempts to reuse that resource on the next test

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: What name do you suggest ? I can change it if you feel it will help.

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
StCostea: I will take a look to see if we can make the driver to not send anything except DISPOSED after FINISHED.

StCostea: I have found a solution to stop sending to the client anything except the DISPOSED message after a FINISHED was sent. But this functionality means that we will have to change the specs also.

I did not commit the changes, as I'm not sure this is the best direction. There is also the option to have a loop in the junit side to read and ignore all events after FINISHED up to DISPOSED. This will not require any spec change.

dpwspoon: I don't understand why the spec says a notify/notified can be sent after a finished.  There isn't any meaning/reason for this that I can find, other than it matched the existing implementation?  So I would suggest a spec change.

StCostea:
You cannot control when a finished is sent. Whenever the script finishes.
In the meantime the junit can send a notify or await barrier. This means
that the notification will never arrive.
We need to think more about the implications to the junit client...

dpwspoon:
I understand that, but the junit shouldn't block on receiving a notified because as you stated, the script execution may already have stopped (i.e. FINISHED is in flight).  I've modified the SPEC here to clarify this case, by saying the driver will only send a notified in the script is still executing (i.e. FINISHED is not sent) and that awaits received during after FINISHED may be ignored (alternatively and Error event could be sent if the await is sent at a completely inappropriate time like after a dispose).

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
dpwspoon: I see, I understand now.  There, was probably an ambiguity in this original code, specifically where script execution has finished.  Specifically, after the prepare/prepared but before started script execution could complete if it was all server channels.  The case here is probably covering that with the notifyPrepared() being called in notifyFinished.

It was unclear to me what the best behavior should be, but we should specify either 

1) the FINISHED event can only come after a START/STARTED,
or 
2) whether it can come after prepared (in case of server channels).

In my PR and what we had originally specified is 1.  But now after thinking about it, perhaps 2 is better because it allows early termination of completed or partially completed and then errored scripts.  Thoughts?

StCostea: notifyPrepared might be needed in case the main thread is in awaitPrepared and there is an abort coming right after preparing the script. Something like this might occur in case we have an assumeTrue or assumeFalse. If we remove it, the main thread will remain blocked.

The FINISHED event can come after an ABORT. And the ABORT can be send even if a START was not sent. We have also the situation when the script contains only servers, but it does not make sense from testing point of view.
The spec is already specifying that the FINISHED can come after an ABORT and an ABORT before the START.
`Any time after the PREPARE command and before DISPOSE, the test framework can send an **ABORT** command. If the execution of the scripts is already finished, no event is sent back. If the execution is aborted by this command, a FINISHED event with the script that was executed so far must be returned.`
So we should still keep it this way.

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
StCostea: Normally, the way a client will implement an AWAIT will be that it will lock a thread waiting for the notification. If this notification will never be sent as a consequence of the already sent FINISHED event, the thread will unlock only with an interrupt (or timeout), which will cause the test to fail, even if it finished on the driver with success.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
dpwspoon: > notifyPrepared might be needed in case the main thread is in awaitPrepared and there is an abort coming right after preparing the script. Something like this might occur in case we have an assumeTrue or assumeFalse.

Prepare Prepared comes before the test execution starts, in the k3po rule with the annotation.  So how would there be multiple threads?   I would assume prepare would block for prepared, then the test method starts getting to the assumetrue/assumefalse

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
dpwspoon: Conversation continued here: https://github.com/k3po/k3po/pull/408#issuecomment-282756168, with clarification to have FINISH include all barriers (state of the world)

dpwspoon:
Looks ok to me, see travis failure, also waiting on SPEC change approval: https://github.com/k3po/k3po/pull/408.  Please also add a comment to the code for this: 

https://github.com/k3po/k3po/pull/406/files/922a996728550233db097250ee53853bdefa6c51#diff-abfce097873151ef6a98d03a67d31d86

on why it is needed.  

Thanks

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: Connection to the driver is done in the additional thread. So prepare also is on that thread. The main thread starts the second thread then waits for prepared notification. Only after prepared the actual test is started. Check SpecificationStatement.evaluate().

PR #407, opened 2017-02-23
Thread leaks in K3po

StCostea:
Branch created from the one for PR 406. Should be reviewed and merged after that PR is merged.

Fixes for Agrona threads not closed after a failed PREPARE command and for worker threads not closed for servers accepting on UDP ports.


driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -371,7 +372,8 @@ public void operationComplete(ChannelFuture future) throws Exception {
 
     @Override
     public void disposeReceived(final ChannelHandlerContext ctx, MessageEvent evt) throws Exception {
-        if (robot == null || robot.getPreparedFuture() == null) {
+//        if (robot == null || robot.getPreparedFuture() == null) {
```
dpwspoon: nit, please just remove old code

driver/src/test/java/org/kaazing/k3po/driver/internal/RobotAgronaThreadsIT.java
```diff
@@ -0,0 +1,100 @@
...
+        public Statement apply(Statement base, Description description) {
+            return new Statement() {
+                @Override
+                public void evaluate() throws Throwable {
+                    Set<Thread> threadsBefore = Thread.getAllStackTraces().keySet();
+                    int threadCountBefore = threadsBefore.size();
+                    base.evaluate();
+
+                    // let the driver shutdown all threads
+                    Thread.sleep(1000);
```
dpwspoon: Is there a way to do this without a sleep.  This is ok but sleeps in tests are bad as total time grows AND it is inherent race when ran on slower build machines.  If not, maybe aggressively try and allow retries for a period of time before failing

driver/src/test/java/org/kaazing/k3po/driver/internal/RobotUdpServerThreadsIT.java
```diff
@@ -0,0 +1,113 @@
...
+        public Statement apply(Statement base, Description description) {
+            return new Statement() {
+                @Override
+                public void evaluate() throws Throwable {
+                    Set<Thread> threadsBefore = Thread.getAllStackTraces().keySet();
+                    int threadCountBefore = threadsBefore.size();
+                    base.evaluate();
+
+                    // let the driver shutdown all threads
+                    Thread.sleep(1000);
```
dpwspoon: same

dpwspoon:
Thanks @StCostea

driver/src/test/java/org/kaazing/k3po/driver/internal/RobotAgronaThreadsIT.java
StCostea: We need to make sure the driver has shut down. I don't know of a way to check it.
I tried without it and sometimes it shows some threads from the driver even if there is a call to `robot.dispose().await();` at the end of the k3po statement, so supposedly the robot is disposed.
I'll take any suggestion if you have one.

StCostea: Do you want me to retry the check several times (with a sleep of, let's say, 50 ms) up to a second or two until `threadCountBefore` equals `threadCountAfter` ?

dpwspoon: > Do you want me to retry the check several times (with a sleep of, let's say, 50 ms) up to a second or two until threadCountBefore equals threadCountAfter ? 

I think that would be better, yes

PR #408, opened 2017-02-24
Suggested control SPEC improvements

dpwspoon:


specification/k3po.control/SPEC.md
```diff
@@ -29,60 +47,20 @@ If the "content-length" header is specified, then a message body of the specifie
...
-This command will initiate a new conversation. Scripts that should be loaded are given in the **name** headers. Additional override values for
-properties existing in the scripts can be given in the message body.
+This **PREPARE** command informs the driver to prepare one or more K3PO script for execution. 
 
 ###### Headers
 - version - the version of the protocol. Currently the only supported version is 2.0.
 - content-length - the length of the message body. This header is optional, should be present only when message body is not empty.
 - origin - the root folder for the scripts to be executed. This header is optional
-- name - the name of the script to be executed. It is a relative path to the origin. Can appear more than once (several scripts to be executed).
+- name - the name of the scripts to be executed. It is a relative path to the origin. Can appear more than once (several scripts to be executed).
```
StCostea: In one 'name' header, there is the name of one script and only one. If you have multiple scripts, the 'name' header will occur more than once, for each of the scripts. So I think that initial version is correct.

specification/k3po.control/SPEC.md
```diff
@@ -114,8 +92,7 @@ N/A
 ##### ABORT command
 
 ###### Description
-This command will abort the current execution. It may be sent any time after a PREPARE command. When the execution of the scripts is aborted,
-a FINISHED event may be returned with the script executed so far.
+The **ABORT** command will abort the currently running script execution.
```
StCostea: ABORT will stop the execution of all scripts in the current test. No matter if a script is running or blocked waiting for an event.

StCostea:
I have discussed with Costin and he came up with an alternate solution. If we can add all barriers and their statuses in the FINISHED events (as we have them in PREPARED), we can make the junit AWAIT be notified by receiving the FINISHED event. In this case we can make the driver send only the DISPOSED after the FINISHED is sent.

As an example:
```
PREPARED
barrier:barrier1
barrier:barrier2
content-length:34
    
connect_url http://localhost:8081
```

```
FINISHED
barrier:barrier1-notified
barrier:barrier2
content-length:34
    
connect_url http://localhost:8081
```
The way to specify that a barrier is notified is just an example, we can come up with a better solution.

specification/k3po.control/SPEC.md
dpwspoon: thanks, fixed

dpwspoon:
@StCostea That sounds like a good plan to me, the finish will need all the completed and incomplete barriers, with the implication being that if a barrier isn't listed then it does not exist in the test.  (i.e. the test framework shouldn't be awaiting it).

dpwspoon:
I would also recommend putting the state in the header name, rather then concat-ing the state to the barrier by convention.  

So my starting proposal would be 
```
completed-barrier: barrier-name
incomplete-barrier: barrier-name
```

specification/k3po.control/SPEC.md
```diff
@@ -240,3 +217,50 @@ This event is sent by the driver when the driver encountered an error while proc
...
+hitting undefined behavior AND 
+  2) after sending a **STARTED** event (note, script execution may have completed if the script
+is just accepts prior to the **START** command or **STARTED** event)
+
+Any time after the **PREPARE** command and before the **DISPOSE** command, the test framework can send an **AWAIT** command.
+ 
+If the driver receives an **AWAIT** command and the scripts are still executing (i.e. the **FINISHED** command has not been sent
+ or is not in the process of being sent), the driver MUST send a **NOTIFIED** event if the barrier has already complete or 
+ when/if the barrier does completes.  The driver MAY ignore await commands if it receives them after script execution.
+
```
dpwspoon: This is resolved by the Finish sending the state of the world on barriers

specification/k3po.control/SPEC.md
```diff
@@ -240,3 +217,50 @@ This event is sent by the driver when the driver encountered an error while proc
...
+Prior to receiving a **FINISHED** command, the test framework may request script execution to stop by sending an  **ABORT** 
+command. The driver must abort the current script execution, which should trigger a **FINISHED** command.  If the script 
+execution has already stopped the driver will ignore the **ABORT** command (note, in this case the driver should already have
+sent the **FINISHED** or be in the process of sending the finished). 
+
+The test framework signals the end of the test by sending a **DISPOSE** command to the client.  Note the test can not end 
+prior to completing the script execution (i.e. having sent a **FINISHED** event).  This triggers the clean up of any
```
dpwspoon: > Also, dispose is executing an abort on the Robot, it is not waiting for the scripts to finish.

I don't believe that is the purpose of dispose nor what I have written here.  I fixed the client/driver typo.  The scripts, successfully completed or not, can leave client or server connections / bindings open.  The dispose is to notify when it safe for those connections to be closed.   I've reworded to reinforce and clarify this meaning.  please review

specification/k3po.control/SPEC.md
```diff
@@ -240,3 +217,50 @@ This event is sent by the driver when the driver encountered an error while proc
...
+Prior to receiving a **FINISHED** command, the test framework may request script execution to stop by sending an  **ABORT** 
+command. The driver must abort the current script execution, which should trigger a **FINISHED** command.  If the script 
+execution has already stopped the driver will ignore the **ABORT** command (note, in this case the driver should already have
+sent the **FINISHED** or be in the process of sending the finished). 
+
+The test framework signals the end of the test by sending a **DISPOSE** command to the client.  Note the test can not end 
+prior to completing the script execution (i.e. having sent a **FINISHED** event).  This triggers the clean up of any 
+outstanding resources on the driver (such as left open server streams).  When all resources have been cleaned up, the driver
+responds with the **DISPOSED** event, the test is then completed and the underlying connection may be reused for another test.
+
```
dpwspoon: Please link to the conversation.  I believe the specification should be written as this, we can file a ticket to implement on the driver side.  I believe our previous conversation was about implementing the reuse of that connection on the client (junit) and deciding there was no need to implement it that way.

specification/k3po.control/SPEC.md
StCostea: DISPOSE can be called anytime. It is like calling an ABORT and then releasing resources. At least this is the current functionality. Also, there can be a situation when there is no FINISHED - when we have an error.

specification/k3po.control/SPEC.md
```diff
@@ -29,60 +47,20 @@ If the "content-length" header is specified, then a message body of the specifie
 
 Headers that are not understood / processed by a command must be ignored.
 
-
-### Flow
-
-The communication between the test framework and the driver is an asynchronous conversation - the test framework sends commands, the driver 
-may respond with one or more events or it may not respond at all (no event sent) when the action is finished.
-
```
dpwspoon: I added an introduction to address this point.  I think what you are getting at here is that it would be a mistake to think of the protocol as request and response.  So I kept the async part and explained events are sent by state changes

specification/k3po.control/SPEC.md
```diff
@@ -192,7 +174,7 @@ N/A
 ##### NOTIFIED event
 
 ###### Description
-This event is sent by the driver when a barrier is triggered.
+This event notifys the client that a barrier has been triggered.  It may be sent after a barrier is triggered, or after a NOTIFY command if the barrier was previously completed.
```
StCostea: The NOTIFY command triggers the barrier. It does not matter if it was completed or not.

specification/k3po.control/SPEC.md
```diff
@@ -139,8 +122,8 @@ N/A
 ##### NOTIFIED command
 
 ###### Description
-This command will trigger on the driver the barrier with the name sent in the header and will have a response with a NOTIFIED event 
-with the same barrier name.
+The **NOTIFIED** command is used by the test framework to unblock script execution that is awaiting a barrier of the same 
+name included in the **NOTIFIED** command.
```
StCostea: You made a confusion between NOTIFY and NOTIFIED. NOTIFY is a command from the test framework, NOTIFIED is an event sent by the driver.
Also, there is nothing in the message exchanges section about NOTIFY command.

specification/k3po.control/SPEC.md
```diff
@@ -240,3 +224,50 @@ This event is sent by the driver when the driver encountered an error while proc
...
+outstanding resources on the driver (such as left open server streams).  When all resources have been cleaned up, the driver
+responds with the **DISPOSED** event, the test is then completed and the underlying connection may be reused for another test.
+
+#### Common Error Cases
+
+Any unrecoverable error case triggers an **ERROR** event.  The underlying connection MUST not be reused.
+
+1. An **ERROR** event is sent if the test framework requests scripts to be prepared that can not be found
+2. An **ERROR **event is sent if the test framework sends an **AWAIT** for a barrier that does not exist
+
```
StCostea: It seems that my comment related to errors is missing. There are a lot of error conditions in the previous version of the document that you did not add. With current version, a lot of functionality is not defined (for example, what happens when you send a second time a START or a PREPARED, if you call any command before calling a PREPARE and others). You can go through previous version and see where/how can you add those error situations.

specification/k3po.control/SPEC.md
```diff
@@ -240,3 +217,50 @@ This event is sent by the driver when the driver encountered an error while proc
...
+when it receives the **START** command.
+
+The driver MUST send the **STARTED** event when the scripts have started.
+
+The driver MUST send the **FINISHED** event, when
+  1) all K3PO scripts have finished, either by successfully completing or 
+hitting undefined behavior AND 
+  2) after sending a **STARTED** event (note, script execution may have completed if the script
+is just accepts prior to the **START** command or **STARTED** event)
+
```
dpwspoon: > The ABORT case is not clearly specified here. I see it bellow, but it can be missed if you don't read carefully the entire paragraph.

I moved that paragraph closer.

> Also, I don't see anywhere the fact that FINISHED is sent maximum one time. In case of an exception running the script we might not even have a FINISHED event, but an ERROR.

Is that not implied?  "Error" is unrecoverable, and "FINISHED" is sent when script execution stops,  that can only be once, correct?

specification/k3po.control/SPEC.md
dpwspoon: swapped NOTIFY with AWAIT

specification/k3po.control/SPEC.md
dpwspoon: I think this was a bug in the original implementation.  I've fixed the definition here.

specification/k3po.control/SPEC.md
dpwspoon: IMHO, the SPEC should specify the meaning of the protocol.  An exhaustive list of every error condition is not really possible, but it should be possible to interpret that the examples you are stating are errors, as they have no meaning.  I've cleaned this up to clearly state, not all error cases are explicitly listed here.

dpwspoon:
I've updated the headers on the *FINISH* commands

dpwspoon:
@StCostea I've tried to address all the feedback, this is ready for another round.

specification/k3po.control/SPEC.md
StCostea: Yes, we decided that way. But the specs and the implementation was somewhere half way. We will have to enforce this in the code and make sure there is no reference to commands after a DISPOSE (like a new PREPARE, for example). I will check one more time the spec tomorrow.

Conversation related to closing connection after an error I think is here: https://github.com/k3po/k3po/pull/390#pullrequestreview-16259036

specification/k3po.control/SPEC.md
StCostea: We should not have an event and a command with the same name. It will generate a lot of confusion in the future. NOTIFY and NOTIFIED is safer.

StCostea:
I have already implemented FINISHED that way. I'm testing it and check the impact on all our tests.

specification/k3po.control/SPEC.md
```diff
@@ -192,7 +174,7 @@ N/A
 ##### NOTIFIED event
 
 ###### Description
-This event is sent by the driver when a barrier is triggered.
+This event notifys the client that a barrier has been triggered.  It may be sent after a barrier is triggered, or after a AWAIT command if the barrier was previously completed.
```
StCostea: NOTIFIED is sent also as a response to NOTIFY, no matter what is the status of the barrier prior to the NOTIFY command.

specification/k3po.control/SPEC.md
```diff
@@ -240,3 +224,54 @@ This event is sent by the driver when the driver encountered an error while proc
...
+If the driver receives an **AWAIT** command and the scripts are still executing (i.e. the **FINISHED** command has not been sent
+ or is not in the process of being sent), the driver MUST send a **NOTIFIED** event if the barrier has already complete or 
+ when/if the barrier does completes.  The driver MAY ignore await commands if it receives them after script execution.
+
+
+The test framework signals the end of the test by sending a **DISPOSE** command to the driver.  Note the test can not end 
+prior to completing the script execution (i.e. DISPOSE MUST not be sent before a **FINISHED** event).  This triggers the clean up of any 
+outstanding resources on the driver (such as left open server streams).  When all resources have been cleaned up, the driver
+responds with the **DISPOSED** event, the test is then completed and the underlying connection may be reused for another test.
+
```
StCostea: After a DISPOSE/DISPOSED, the connection cannot be reused. A new connection must be opened for a new test. It is cleaner and with more separation between two tests.

specification/k3po.control/SPEC.md
dpwspoon: > NOTIFIED is sent also as a response to NOTIFY, no matter what is the status of the barrier prior to the NOTIFY command.

It isn't sent if the barrier is not yet trigger

dpwspoon: Oh I see, thanks!

dpwspoon: updated

specification/k3po.control/SPEC.md
```diff
@@ -8,75 +8,58 @@ This document defines the K3po control protocol used by a test framework to comm
...
+- command              = token
+- header-field = field-name ":" field-value
+- LF, message-body, field-name, field-value, and token are defined in [HTTP RFC](https://tools.ietf.org/html/rfc7230).
 
-- Any time after the PREPARE command, the test framework can send a **DISPOSE** command. The driver must abort the execution and release the 
-resources in use. When done, a DISPOSED event must be sent back. At this point you can consider that the test has ended and the driver is back in the initial state.
+If the "content-length" header is specified, then a message body of the specified length must be added.
 
-- After a DISPOSE command, no other command is accepted except a new PREPARE. Any other command will have as a result an error with a message "Script has not been prepared or is still preparing". 
+Headers that are not understood / processed by a command must be ignored.
```
jfallows: The rationale for relaxed input processing is to support independent upgrade of server and client, however we always upgrade the k3po server at (say) k3po control client at the same time in the project dependencies, so it would be better to be strict on unrecognized headers.

specification/k3po.control/SPEC.md
```diff
@@ -216,11 +198,13 @@ N/A
 ##### FINISHED event
 
 ###### Description
-This event is sent by the driver when the driver has finished execution of the scripts. Only one FINISHED event is sent for a script execution.
+This event is sent by the driver when the driver has finished execution of the scripts. 
 
 ###### Headers
 - content-length - the length of the message body. This header is optional, should be present only when the message body is 
 not empty.
+- notified-barrier - (occurs 0 or more times), the value is the name of a barrier that completed.  All completed barriers MUST be sent in the FINISHED event.
```
jfallows: `notified: barrier-name`

specification/k3po.control/SPEC.md
```diff
@@ -216,11 +198,13 @@ N/A
...
+- await-barrier -  (occurs 0 or more times), the value is the name of a barrier that did not complete.  All incomplete barriers must be sent in the FINISHED event.
```
jfallows: `awaiting: barrier-name`

specification/k3po.control/SPEC.md
jfallows: Can we eliminate the need for `DISPOSE/DISPOSED` round trip and instead signal resource cleanup by closing the control connection?

This must already be required to properly cleanup in cases where the control connection is severed unexpectedly.

dpwspoon: The `DISPOSE/DISPOSED` is there because some scripts leave the network in a state that can break the next junit test if there is a race between finishing clean up and start of next test.  For a hypothetical example consider the script,

bind.on.8000.rpt
```
accept tcp://localhost:8000
accepted
```

and the junit test class with

```java

@Test
public void connectTo8000(){
 // Client Connects
}

@Test
public void bindTo8000(){
 // Server Binds to 8000
}
```

After the first test `connectTo8000`, if the driver is unbinding and `bindTo8000` starts, then it is possible the second test might fail.

I don't remember the specifics, but we have seen this in tests in the past, and introduced DISPOSE/DISPOSE as a way to wait for clean up. 

Another solution we could consider is to have the driver not run scripts in parallel, i.e. don't return the subsequent prepared until scripts before were cleaned up.  I believe we chose `dispose/disposed` cause we were still thinking about supporting that use case??

jfallows: It would be much better to make the server resilient to completing the resource cleanup of the previous test before preparing the next one, and keep the control protocol simple.

specification/k3po.control/SPEC.md
dpwspoon: > The rationale for relaxed input processing is to support independent upgrade of server and client, however we always upgrade the k3po server at (say) k3po control client at the same time in the project dependencies

This is not necessarily true with javascript/c/dotnet etc clients

specification/k3po.control/SPEC.md
dpwspoon: > It would be much better to make the server resilient to completing the resource cleanup of the previous test before preparing the next one, and keep the control protocol simple.

Sounds good to me.

specification/k3po.control/SPEC.md
jfallows: We shouldn't leave them behind with old control protocol if they need to use a more modern version of the k3po server - that's part of the upgrade in some cases.

dpwspoon:
All feedback applied

specification/k3po.control/SPEC.md
jfallows: However, if there is an incompatible semantic change, let's track that by updating the `version` header in `PREPARE` as needed.

When protocol frames are produced, they writer must be precisely accurate, but when protocol frames are consumed, the reader should be tolerant of additional unrecognized headers on recognized frame types.

specification/k3po.control/SPEC.md
```diff
@@ -50,7 +50,7 @@ are formatted as follows:
 
 If the "content-length" header is specified, then a message body of the specified length must be added.
 
-Headers that are not understood / processed by a command must be ignored.
+Headers that are not understood / processed must result in an ERROR event if received by the server, or a closed connection if received on the client
```
jfallows: We probably need to relax this a little bit, so that writers of the protocol are precise and readers of the protocol are tolerant.

specification/k3po.control/SPEC.md
```diff
@@ -50,7 +50,7 @@ are formatted as follows:
 
 If the "content-length" header is specified, then a message body of the specified length must be added.
 
-Headers that are not understood / processed must result in an ERROR event if received by the server, or a closed connection if received on the client
+Headers and messages that are not understood/processed by a receiver may be ignored. The writers must only send headers and messages that it understands.
```
jfallows: `should` be ignored?

specification/k3po.control/SPEC.md
```diff
@@ -50,7 +50,7 @@ are formatted as follows:
 
 If the "content-length" header is specified, then a message body of the specified length must be added.
 
-Headers and messages that are not understood/processed by a receiver may be ignored. The writers must only send headers and messages that it understands. 
+Headers and messages that are not understood/processed by a receiver should be ignored. The writers must only send headers and messages that it understands.
```
jfallows: Please change this to the following for precision (in general we should be following RFC convention and using SHOULD, MUST, MUST NOT, etc rather than should, must, must not, etc for specification content)

Receivers SHOULD ignore undefined message headers and message types. Senders MUST NOT send undefined message headers or message types.

dpwspoon: fixed

PR #410, opened 2017-03-06
Sporadic failure in ConnectionManagementIT

StCostea:
Test ConnectionManagementIT#clientWithPipeliningMustNotRetryPipeliningImmediatelyAfterFailure fails in Ubuntu but is successful under Windows.

### DIFF
--- a/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/client.with.pipelining.must.not.retry.pipelining.immediately.after.failure/request.rpt
+++ b/specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/client.with.pipelining.must.not.retry.pipelining.immediately.after.failure/request.rpt
@@ -29,6 +29,7 @@ read "HTTP/1.1 200 OK\r\n"
 read "Content-Length: 8\r\n"
 read "\r\n"
 read "request1"
+closed
 read notify REPLAY_REQUEST_2
 
 connect await REPLAY_REQUEST_2


specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/client.with.pipelining.must.not.retry.pipelining.immediately.after.failure/request.rpt
```diff
@@ -29,6 +29,7 @@ read "HTTP/1.1 200 OK\r\n"
 read "Content-Length: 8\r\n"
 read "\r\n"
 read "request1"
+closed
```
dpwspoon: Why not use a `read closed` to keep it in the http layer.  Closed will wait for tcp layer to close, I think?? also, the ordering looks weird to me, could we put the read notify above the read close or, tie it to the closed (`closed notify` this might not be implemented yet?)?

StCostea: That is exactly the issue in this bug. The server has two different responses on `http://localhost:8080/request2`, but the order is messed up on Ubuntu, the first answer goes to the second request from the client script. So, we need to make sure that the first connection is closed (and the server sent the response for the first request2) and then make the second request. If we will put the notify before the close, the close will be useless and the functionality will be the same as before the change.

I'm not sure what is the functionality of `read closed`, if it is equivalent to `closed`. I'll check it and add another comment here.

StCostea: `read closed` can be used in a http script. The client script is a tcp script, so we cannot use it. Also, `read closed` does not guarantee that the connection is closed. It just signals to the driver that the request is ended.
From the k3po wiki:
> In order to inform the K3PO driver that no more headers are expected and that it is alright to flush the request to the wire, a read close or write close is expected after the request or response.
Nothing about closing the connection.

StCostea: If you want, I can remove the `read`, just leave `notify REPLAY_REQUEST_2`

Anisotrop:
This solution seems to function. It hides however a difference in the accept behavior at tcp and http level. An http accept does not distinguish between a request on a new tcp connection or a request reusing an existing connection, when declaring accepts on same host:port like the following:
```
accept http://localhost:8080/request1
accept http://localhost:8080/request2
```

An accept option that requires the accept to be on the same connection might solve the issue but could be difficult to implement and also might require similar logic on the connect part of http.

Another option would be to implement the server script also as a tcp script with control over the connections like [this PR](https://github.com/k3po/k3po/pull/411)

specification/http/src/main/scripts/org/kaazing/specification/http/rfc7230/connection.management/client.with.pipelining.must.not.retry.pipelining.immediately.after.failure/request.rpt
dpwspoon: What you have here looks good to me

dpwspoon: Thanks for the background

PR #241, opened 2015-10-23
Added Dispose/Disposed to command and event protocol to clean up k3po driver

dpwspoon:
Resolves https://github.com/k3po/k3po/issues/239

Issue #239, opened 2015-10-20
Add new command to k3po to execute clean up from junit rule.


dpwspoon:
I believe the assumption when creating K3po was that the cleanup of K3po was executed in k3po.finish() call.  However this is not exactly correct

Currently we call Robot.destroy() from the ControlServerHandler channelClosed

```java
    @Override
    public void channelClosed(ChannelHandlerContext ctx, ChannelStateEvent e) throws Exception {
        if (robot != null) {
            robot.destroy();
        }
        channelClosedFuture.setSuccess();
        ctx.sendUpstream(e);
    }

```

K3po.destroy releases the external resources used by the bootstrap factory

```java
    public boolean destroy() {

        if (destroyed) {
            return true;
        }

        abort();

        if (createdBootstrapFactory) {
            try {
                bootstrapFactory.shutdown();
                bootstrapFactory.releaseExternalResources();
            }
            catch (Exception e) {
                if (LOGGER.isDebugEnabled()) {
                    LOGGER.debug("Caught exception releasing resources", e);
                }
                return false;
            }
        }

        return destroyed = true;
    }
```

I believe this may cause a race between the release of external resources and the starting of a next test, which could manifest itself like this https://github.com/k3po/k3po/issues/229

We could clean up the bound channel resources on the k3po.finish (releaseExternalResources there), but logically I think this would be better done when the rule chain is returning.

I propose synchronizing the shutdown of k3po driver with a new command and event.
For now we will call these
`TERMINATE` and `TERMINATED` respectively.  (These are okay, looking for something better if there is something)

Terminate will be sent from the K3poRule after the test logic has executed and is returning out of the chain (perhaps in the verify() call)

dpwspoon:
Additionally, if clean up of a k3po accept is done at K3po.finish, this can causes junit code to observer behavior not specificed in the script (such as a close).  Meaning an assert or mock might have unexpected behavior

dpwspoon:
reviewed with @jfallows.  Sounds good, though perhaps we could have better verbs then terminate and terminated.  Perhaps clean and cleaned.  Terminate is sort of alright though cause you are terminating the k3po.driver class (K3po)

cmebarrow:
@dpwspoon Are you saying connections established by the Robot (via connect and accept) will remain open unless explicitly closed in scripts until terminate (or clean) is sent? I think that will resolve many sporadic test failures we've seen, including in tests like the jms tests that use mock objects, as you pointed out above.

Issue #229, opened 2015-10-12
httpxe specification tests notably in RequestsIT and OriginSecurityIT sometimes fail due to failed connect stream


cmebarrow:
See for example [k3po Travils build #332.1](https://travis-ci.org/k3po/k3po/jobs/84894172). The test failed as follows:
```
Running org.kaazing.specification.httpxe.RequestsIT
Tests run: 17, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.989 sec <<< FAILURE! - in org.kaazing.specification.httpxe.RequestsIT
shouldProcessPathEncodedPostRequestAsOptionsRequest(org.kaazing.specification.httpxe.RequestsIT)  Time elapsed: 0.07 sec  <<< FAILURE!
org.junit.ComparisonFailure: Specified behavior did not match expected:<[connect http://localhost:8000/;options/path
connected

write method "POST"
write version "HTTP/1.1"
write header content-length
write flush]

accept http://loca...> but was:<[]

accept http://loca...>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.kaazing.k3po.junit.rules.SpecificationStatement.evaluate(SpecificationStatement.java:127)
	at org.junit.rules.Verifier$1.evaluate(Verifier.java:35)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
```

mgherghe:
There are also other httpxe tests failing on Travis builds. See builds: [#319.1](https://travis-ci.org/k3po/k3po/jobs/84051138),  [#319.2](https://travis-ci.org/k3po/k3po/jobs/84051141), [#322.2](https://travis-ci.org/k3po/k3po/jobs/84279223), [#324.1](https://travis-ci.org/k3po/k3po/jobs/84325616), all with the same error message as above.

cmebarrow:
Same failure occurred multiple times in [this](https://travis-ci.org/cmebarrow/k3po/builds/90592278) Travis build, console output is available at:
[build33_diagnostics_httpxe_fail_k3o_issue_229.txt](https://files.zenhub.io/5643ae4ee5b473fe074acef2)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

PR #417, opened 2017-03-21
Removal of DISPOSE and DISPOSED messages.

StCostea:


dpwspoon:
This looks good to me, but it needs to go in at the same time with the logic to not start the next test if an existing test is still cleaning up.  I.E. clean up on socket closed BUT also don't process PREPARE (hold on to it) until clean up is complete.  Otherwise there is a race

StCostea:
A new socket connection will use a new Robot. So there is no race, as there are different instances of the k3po engine.
Also, it is not possible to reuse a connection for a second test.

dpwspoon:
> A new socket connection will use a new Robot. So there is no race, as there are different instances of the k3po engine.

What about finite resources, such as binding to ports?  Only one robot can use one at a time, so the other needs to be cleaned up prior to the other starting.  This is why disposed was introduced originally, to eliminate that race.

StCostea:
I guess you are talking about port 11642 for commands. It is outside the
robot, it is only one instance listening. It is not affected by dispose. As
for ports used in the scripts, they will conflict anyway if you run two
scripts in parallel, which you can do even now.
I don't know of any other resources that can be shared between different
instances of the robot.

And, by the way, when the connection is closed, we still have the call to
robot.dispose.

dpwspoon:
I'm talking about the ports in the scripts themselves, say `accept tcp://localhost:8000` is used by two tests.

1. Test 1 starts and binds to 8000 in Robot instance 1
2. Test 1 completes, kills control channel connection, Robot 1 starts to clean up
3. Test 2 starts via Robot instance 2, tries to bind to 8000 and fails
4. Robot 1 finishes clean up.

jfallows:
We need to defer processing of the next `PREPARE` until after the previous control connection has closed and associate resources have been cleaned up, right?
That will prevent any resources needed by next `PREPARE` from colliding with the resources used by previous test scenario, including port collisions across tests.

dpwspoon:
Yes

StCostea:
Currently, the dispose will be called when the connection closes. We can force the call to dispose when a PREPARE is received on another connection. 

If we change it like this, we will always support only one connection to the driver. If connection is not closed (or at least the execution finished), we will not be able to accept another connection. I don't think it is ok to force close connections after FINISHED.

jfallows:
Let's summarize the requirements
- driver starts listening for control connections before all ITs
- test framework uses a different control connection to driver for each IT test method
- resources needed by each IT test method (eg. bind ports) must be available
- driver stops listening for control connections after all ITs

Given the potential issue with bind port collision if test resources are not cleaned up in time, we need to queue the processing of a `PREPARE` until resources used by the previous test have been cleaned up.

There appears to be no need to aggressively close control connections to manage test resources properly.

StCostea:
The test resources were cleaned by the DISPOSE command. Without it, the only way to trigger cleaning them is with the closing of the connection. If the client does not close connection, they will be cleaned only after a tcp connection timeout. If we will add a wait for resources to be cleared, all tests will have to close connection or we will have a big slow down of integration tests.

dpwspoon:
> If the client does not close connection,

The client is k3po JUnit.  We should make it behave correctly and close the control connection for all the relevant cases.  The unusual case we need to watch out for are when the test fails, but even then I would expect we can program it correctly, and test failures should be the rare case.  Thus the timeout should not occur with a well-behaved test framework.

Eventually, if we got programmatic access to the closing handshake on TCP, we could even have that replace the dispose/disposed (first fin is dispose and sent by client, second fin is the disposed).  I don't think Netty 3.x will support that but when we migrate to reaktivity we will be able to do that.  I will file a follow-up ticket after the current work is complete.

StCostea:
I think I have found some tests failing that seem to be from this cause. After I will finish with the gateway issue, I will try to make sure that the client closes the connection, and only then moves on to another test.

StCostea:
I have run several test runs on k3po and gateway, it seems that the port binding issue does not occur.

I also checked the code - on ending the test, the specification will call `scriptRunner.dispose()` in a finally block and the dispose method will close the connection. So I don't think there is any code change required. The current version will not have any issues with finite resources as long as the tests run one at a time. And this is what is happening in any maven build.

Also, after the log rule bug is fixed, the build for gateway runs in a loop for several hours without any errors. So, I consider that this PR is ready for merge.

dpwspoon:
I see this new exception in local builds with this PR:

```
Running org.kaazing.specification.wse.ClosingIT
[WARNING] EXCEPTION, please implement org.kaazing.k3po.driver.internal.Robot$CloseOnExceptionHandler.exceptionCaught() for proper handling.
java.nio.channels.ClosedChannelException
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker.cleanUpWriteBuffer(AbstractNioWorker.java:433)
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker.writeFromUserCode(AbstractNioWorker.java:128)
	at org.jboss.netty.channel.socket.nio.NioClientSocketPipelineSink.eventSunk(NioClientSocketPipelineSink.java:84)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendDownstream(DefaultChannelPipeline.java:779)
	at org.jboss.netty.channel.Channels.write(Channels.java:725)
	at org.jboss.netty.handler.codec.oneone.OneToOneEncoder.doEncode(OneToOneEncoder.java:71)
	at org.jboss.netty.handler.codec.oneone.OneToOneEncoder.handleDownstream(OneToOneEncoder.java:59)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendDownstream(DefaultChannelPipeline.java:591)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendDownstream(DefaultChannelPipeline.java:784)
	at org.jboss.netty.channel.SimpleChannelHandler.writeRequested(SimpleChannelHandler.java:292)
	at org.kaazing.k3po.driver.internal.netty.bootstrap.http.HttpChannelHandler.httpWriteRequested(HttpChannelHandler.java:94)
	at org.kaazing.k3po.driver.internal.netty.bootstrap.http.HttpChannelHandler.writeRequested(HttpChannelHandler.java:59)
	at org.jboss.netty.channel.SimpleChannelHandler.handleDownstream(SimpleChannelHandler.java:254)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendDownstream(DefaultChannelPipeline.java:591)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendDownstream(DefaultChannelPipeline.java:582)
	at org.jboss.netty.channel.Channels.write(Channels.java:704)
	at org.jboss.netty.channel.Channels.write(Channels.java:671)
	at org.jboss.netty.channel.AbstractChannel.write(AbstractChannel.java:348)
	at org.kaazing.k3po.driver.internal.netty.bootstrap.http.HttpClientChannelSink.shutdownOutputRequested(HttpClientChannelSink.java:320)
	at org.kaazing.k3po.driver.internal.netty.bootstrap.http.HttpClientChannelSink.closeRequested(HttpClientChannelSink.java:296)
	at org.kaazing.k3po.driver.internal.netty.bootstrap.channel.AbstractChannelSink.eventSunk(AbstractChannelSink.java:39)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendDownstream(DefaultChannelPipeline.java:779)
	at org.jboss.netty.channel.SimpleChannelHandler.closeRequested(SimpleChannelHandler.java:334)
	at org.jboss.netty.channel.SimpleChannelHandler.handleDownstream(SimpleChannelHandler.java:260)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendDownstream(DefaultChannelPipeline.java:591)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendDownstream(DefaultChannelPipeline.java:582)
	at org.jboss.netty.channel.Channels.close(Channels.java:812)
	at org.jboss.netty.channel.AbstractChannel.close(AbstractChannel.java:206)
	at org.kaazing.k3po.driver.internal.Robot$CloseOnExceptionHandler.exceptionCaught(Robot.java:552)
	at org.jboss.netty.channel.SimpleChannelHandler.handleUpstream(SimpleChannelHandler.java:130)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:559)
	at org.jboss.netty.channel.Channels.fireExceptionCaught(Channels.java:525)
	at org.kaazing.k3po.driver.internal.netty.bootstrap.http.HttpClientChannelSource.channelClosed(HttpClientChannelSource.java:175)
	at org.jboss.netty.channel.SimpleChannelHandler.handleUpstream(SimpleChannelHandler.java:106)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendUpstream(DefaultChannelPipeline.java:791)
	at org.jboss.netty.handler.codec.replay.ReplayingDecoder.cleanup(ReplayingDecoder.java:570)
	at org.jboss.netty.handler.codec.frame.FrameDecoder.channelClosed(FrameDecoder.java:371)
	at org.jboss.netty.handler.codec.http.HttpClientCodec$Decoder.channelClosed(HttpClientCodec.java:221)
	at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:88)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:559)
	at org.jboss.netty.channel.Channels.fireChannelClosed(Channels.java:468)
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker.close(AbstractNioWorker.java:375)
	at org.jboss.netty.channel.socket.nio.NioWorker.read(NioWorker.java:93)
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker.process(AbstractNioWorker.java:108)
	at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:337)
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker.run(AbstractNioWorker.java:89)
	at org.jboss.netty.channel.socket.nio.NioWorker.run(NioWorker.java:178)
	at org.jboss.netty.util.ThreadRenamingRunnable.run(ThreadRenamingRunnable.java:108)
	at org.jboss.netty.util.internal.DeadLockProofWorker$1.run(DeadLockProofWorker.java:42)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

```

This is not causing the test to fail.

dpwspoon:
You can force the race with a breakpoint.  

### To reproduce, for specification.wse:

Set a break point on line 116 of HttpServerChannelSink and remote debug into the k3po driver:  `mvnDebug k3po:start -Dmaven.k3po.daemon=false -X`  (Note, I had to change the debug port so it didn't conflict with the tests.   

Then from eclipse run BrowserIT (any IT should work).  First, test passes, all other tests fail with address already in use:

image

I highly suspect that if you set the breakpoint further in the stack such that it partially unbounds, then you will be able to reproduce the exact error that dispose/disposed was originally introduced for:

[Dispose PR](https://github.com/k3po/k3po/pull/241)
[Dipose Ticket](https://github.com/k3po/k3po/issues/239)


[One of the tests that failed prior to PR and then closed after merge of dispose PR](https://github.com/k3po/k3po/issues/229),  With log

```
accept http://loca...> but was:<[]

accept http://loca...>
    at org.junit.Assert.assertEquals(Assert.java:115)
    at org.kaazing.k3po.junit.rules.SpecificationStatement.evaluate(SpecificationStatement.java:127)
    at org.junit.rules.Verifier$1.evaluate(Verifier.java:35)
    at org.junit.rules.RunRules.evaluate(RunRules.java:20)
```

indicating a failure on binding where another test was binding.

This case was very rare and very hard to reproduce (I'm not sure if I ever got it to reproduce locally).  

I suspect, the exact log above was perhaps where the server socket went in a Time-Wait state for a connection, unbound, but then can't bind again until the Time-Wait connection times out.

- 

This is a good PR, but it will have a minor regression for this very unlikely bug...  If I recall correctly, the original issue did occur occasionally across tests in kaazing/gateway, and it took us quite some time to track down the cause (i.e. it wasn't an isolated issue to the WSE specification failure linked to the ticket, it was in at least a few other places in kaazing/gateway)

StCostea:
Using the breakpoint is actually blocking the first test thread. It is normal that the second one will fail (unless you really want to make a PREPARE wait until the other test is finishing the clean-up). But this will never happen in normal execution.

I also checked the PR you did for DISPOSE. You did changes in the Robot and also in the communication between the junit and the driver. What I did was only to remove the communication part. The code in the Robot is still there. The only difference is that it is triggered by the closing of the connection and not by the DISPOSE command. So I would expect it to work better after my changes than before your changes.

As for the time-wait state, it will not occur as long as the client is the one that is requesting the closure of the connection. Which is what is currently happening.

The only chance that an error occurs is when a test ends the connection, a new one starts, sends the PREPARE and the PREPARE is processed by the driver, all this **before** the previous test manages to unbind the ports. I consider the chance to get to this situation to be extremely small. Only reading the new scripts from the disc will give more than enough time for the other thread to unbind the ports.

StCostea:
Ok. I will do the changes.

StCostea:
David, how did you do the test with the breakpoint in HttpServerChannelSink ? I tried it and the result is different. The first test passes and stops at the breakpoint. The others get stuck and end with a timeout. I added a breakpoint in prepareReceived and no other test reaches that point.
The problem, as I see it, is that there is only one worker thread in k3po driver and that one I stopped with the breakpoint. So there is no thread available to process the events for the rest of the tests. I don't see how did you get the rest of the tests processed.

This is causing me problems because I cannot just block waiting in prepareReceived for the previous test to end, as it needs the thread available to process the closing of the previous session/robot. I need to somehow delay the processing of the PREPARE event until close connection event is processed for the previous test.

StCostea:
I did a test with keeping the previous robot and just wait for disposeFuture to be done. It works in my tests. I have seen that closing is done on a separate thread after abort() is finished.
I have added the following code in `prepareReceived`:
```
        if (lastRobot != null && ! lastRobot.getDisposedFuture().isDone()) {
            // we need to wait for it to end
            do {
                System.out.println("previous did not end " + System.identityHashCode(this));
                Thread.sleep(100);
            } while (! lastRobot.getDisposedFuture().isDone());
        }
```
A problem would occur if the close connection event is not processed before the prepare will come on the new connection. In that case we will be stuck in the loop waiting. Do you have a suggestion for that case ?

dpwspoon:
There are two running processes: 

***1) The driver with the debugger.*** 

Go to `k3po/specification/wse`, run `mvnDebug k3po:start -Dmaven.k3po.daemon=false -X` with `MAVEN_DEBUG_OPTS="-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=4444"` (note: my MAVEN_DEBUG_OPTS are hardcoded and not set able from the environment in my mvnDebug shell script, I'm not sure if I originally changed that or it came that way).  Remote bind to it

image.

And set break point at

image

image

Run BrowserIT with out debug from a new test:

image

jfallows:
@StCostea the code block in `RobotServer.prepareReceived` at lines 152-163 (shown below) needs to be deferred until all previous robots have finished cleaning up, no special cases for the last robot only.

```
if (scriptLoader != null) {
    Thread currentThread = currentThread();
    ClassLoader contextClassLoader = currentThread.getContextClassLoader();
    try {
        currentThread.setContextClassLoader(scriptLoader);
        prepareFuture = robot.prepare(aggregatedScript);
    } finally {
        currentThread.setContextClassLoader(contextClassLoader);
    }
} else {
    prepareFuture = robot.prepare(aggregatedScript);
}
```

The `preparedFuture` should still be visible outside this block, even though it is potentially deferred, but the `preparedFuture` should only complete when both the previous robots are all cleaned up and the future returned by `robot.prepare` has also completed.

Care must be taken to both execute the `robot.prepare` method on the correct thread when previous cleanup completes, and also to propagate failure if the future completes in error.

This non-blocking approach avoids `Thread.sleep(...)` and will allow the I/O thread to make progress as needed.

StCostea:
@jfallows I'll look into it more tomorrow. Thanks.

StCostea:
@jfallows `all previous robots have finished cleaning up, no special cases for the last robot only` - if we are waiting for previous robots to end, there will always be a maximum of one previous robot. So last robot is not a special case, it is the only case we have.
As an example:
-test A is the first test, there is no previous robot, so it starts
-test B checks that there is test A not yet finished and will wait for it to end before starting
-test C checks that there is test B not yet finished (test A is already stopped when B starts), wait for B to stop and then starts
And so on. 

You need to keep in mind that there is only one I/O thread in the k3po driver, so there cannot be two tests starting in parallel. Also, maven is executing tests sequentially, we do not have two requests for PREPARE at the same time. The second one will come only after the connection for the first one is closed. We only have to make sure that the next test will start after the full processing of the close connection event for the previous test.

StCostea:
@jfallows 
>Care must be taken to both execute the robot.prepare method on the correct thread when previous cleanup completes
This is the part that I don't know how to do. In `Robot.dispose()` there is a `new Thread(new Runnable())` that is executing the code I try to add to Robot.getDisposedFuture().addListener(). So I need to get the MessageEvent to be processed on the IO thread. But I cannot figure how.

dpwspoon:
Hypothetically assume there is two or more JUnit process running in parallel and only one k3po driver.  Maybe someone is running tests in their IDE.  In this case, the two JUnit tests could both send a prepare when a previous k3po test is still cleaning up.  In one design, the two tests will wait for the "1" robot to finish and then start and maybe collide.  

A better more generic solution is to queue the tests, sending the Prepared after the one above it in the queue has finished.

So it is important to consider the case where there is more than one k3po above it, and then if you solve that generically, you solve for the general but special case that there is only one.

> A problem would occur if the close connection event is not processed before the prepare will come on the new connection. In that case we will be stuck in the loop waiting. Do you have a suggestion for that case ?

Use futures.  Hook into the disposed future of the robot currently running (or something similar if it does not exist).  Maybe create a queue of robots waiting on it.  On completion pop the queue until you find a robot that is still open.  Then process that guy.

StCostea:
I was talking about the driver having only one worker thread. It does not matter how many PREPARE event are coming in parallel. It is only one thread in the driver so they will be processed sequentially. You cannot process simultaneously on one thread. 

We have these lines in RobotServer:
```
        NioClientBossPool clientBossPool = new NioClientBossPool(newCachedThreadPool(), 1);
        NioServerBossPool serverBossPool = new NioServerBossPool(newCachedThreadPool(), 1);
        NioWorkerPool workerPool = new NioWorkerPool(newCachedThreadPool(), 1);
````

**Comment by StCostea** [comment]
I have managed to find a solution. I have committed a draft for you to see. I still need to test it with full builds.

File: driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java [hunk 1]
```diff
@@ -105,6 +109,29 @@ public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) t
...
+            lastRobot.getDisposedFuture().addListener(new ChannelFutureListener() {
+                @Override
+                public void operationComplete(ChannelFuture future) throws Exception {
+                    ((NioSocketChannel)ctx.getChannel()).getWorker().executeInIoThread(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
+                                prepareReceived(ctx, evt);
+                            } catch (Exception e) {
+                                // TODO Not sure what needs to be done here.
```
dpwspoon: If the connection is still open it should send an Error message to the client

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -105,6 +109,29 @@ public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) t
             return;
         }
 
+        if (lastRobot != null && ! lastRobot.getDisposedFuture().isDone()) {
+            // we need to wait for it to end
+            lastRobot.getDisposedFuture().addListener(new ChannelFutureListener() {
+                @Override
+                public void operationComplete(ChannelFuture future) throws Exception {
```
dpwspoon: This is ok, but it might be cleaner and more future proof to write it without a dependency on the underlying thread model.  I.E. solve for the case where there are more than one thread

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -106,6 +109,28 @@ public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) t
             return;
         }
 
+        if ((lastRobot != null && ! lastRobot.getDisposedFuture().isDone()) || ! testFinished.compareAndSet(true, false)) {
```
dpwspoon: I think it is better the way it was before without the testFinished.compareAndSet.  Seems to be added complexity.  Alternatively, you could reduce the lastRobot and testFinished to a single variable most likely with a a queue

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -66,6 +66,8 @@
     private static final String ERROR_MSG_ALREADY_PREPARED = "Script already prepared\n";
     private static final String ERROR_MSG_ALREADY_STARTED = "Script has already been started\n";
 
+    private static Robot lastRobot;
+    private static AtomicBoolean testFinished = new AtomicBoolean(true);
```
dpwspoon: same nit on the static here.

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -106,6 +109,28 @@ public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) t
...
+            lastRobot.getDisposedFuture().addListener(new ChannelFutureListener() {
+                @Override
+                public void operationComplete(ChannelFuture future) throws Exception {
+                    ((NioSocketChannel)ctx.getChannel()).getWorker().executeInIoThread(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
+                                prepareReceived(ctx, evt);
+                            } catch (Exception e) {
+                                sendErrorMessage(ctx, e);
```
dpwspoon: Can we get a test for this

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -66,6 +66,8 @@
     private static final String ERROR_MSG_ALREADY_PREPARED = "Script already prepared\n";
     private static final String ERROR_MSG_ALREADY_STARTED = "Script has already been started\n";
 
+    private static Robot lastRobot;
```
dpwspoon: Suggest renaming last robot to "currentlyExecutingRobot" or "executingRobot"

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -65,6 +68,7 @@
...
```
StCostea: We have one ControlServerHandlers instance per connection. I have checked it in debug. So we need the static. 
If there was only one instance, everything will be messed up with multiple connections, as there are many non-static fields in the class. For example, channelClosedFuture will be completed by the first test and will be set from the start for all other tests.

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
StCostea: I have tried to see how the old k3po handles an exception in prepareReceived. It will not respond with any error message and the test times out. So I included everything in a try-catch and removed the `throws Exception` statement. So this piece of code is not required anymore. Please see changes.

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
StCostea: Adding a synchronization between threads for sure is adding complexity. You cannot have without the other.

Regarding the queue, I guess you mean to keep the requests to be processed in a queue. I also need the dispose future for the current executing robot. I don't know how to have a Robot and a list of Runnable-s or PrepareEvent-s in one queue and then remove a Robot and a Runnable but add at the beginning of the queue another Robot. I don't even know what kind of structure can I use for this.

I tried also to think how to replace the atomic boolean with a queue. I don't know how to guarantee that a new test coming exactly when the previous one finishes will be executed. I still need some kind of synchronization which will add more complexity.

If you have any suggestion on how to do it, please let me know.

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
dpwspoon: Then please pass it into the constructor

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
```diff
@@ -99,81 +103,70 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
     }
 
     @Override
-    public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) throws Exception {
-        if (robot != null && robot.getPreparedFuture() != null) {
-            sendErrorMessage(ctx, ERROR_MSG_ALREADY_PREPARED);
-            return;
-        }
+    public void prepareReceived(final ChannelHandlerContext ctx, MessageEvent evt) {
+        try {
```
dpwspoon: consider moving try catch to calling class

driver/src/main/java/org/kaazing/k3po/driver/internal/control/handler/ControlServerHandler.java
dpwspoon: You can probably create a queue by just chaining disposedFutures and have it very similar to what you have now: something like 
```
 public synchronized void schedulePrepare(){
        Robot robot = new Robot();  // I think this is already done outside the method
        if(queueTail == null){
            startNow();
        }else{
            queueTail.get().addListener(new ChannelFutureListener() {
                @Override
                public void operationComplete(ChannelFuture future) throws Exception {
                    startNow();
                }
            });
        }
        queueTail.set(robot.getDisposedFuture());
    }
```

jfallows:
@StCostea i think you are on the right track with using Atomics, but suggest the following changes.

Pass an `AtomicReference<Robot>` for `activeRobotRef` to `ControlServerHandler`.
```
public ControlServerHandler(AtomicReference<Robot> activeRobotRef) {
    this.activeRobotRef = activeRobotRef;
}
```

Remove `synchronized` from `prepareReceived`.
In `prepareReceived` use a CAS operation to set the `activeRobot` atomically, and reschedule the next attempt to process `prepareReceived` when the race is lost.
```
@Override
public void prepareReceived(ChannelHandlerContext ctx, MessageEvent evt) throws Exception {
...
if (robot == null) {
    robot = new Robot();
}

if (activeRobotRef.compareAndSet(null, robot)) {
    // success, proceed as before with prepareReceived logic
}
else {
    Robot activeRobot = activeRobotRef.get();
    if (activeRobot == null) {
        channel.getPipeline().execute(() -> prepareReceived(ctx, evt));
    }
    else {
        ChannelFuture disposeFuture = activeRobot.getDisposeFuture();
        disposeFuture.addListener(f ->  prepareReceived(ctx, evt));
    }
}
```

I'm looking for the least complex solution to meet the requirement, let me know if you think this works.

Note: the `activeRobot == null` check is only necessary to cover the more general case with more than one I/O thread, but worth keeping to avoid the introduction of subtle bugs later if we go back to having more than one I/O thread for the RobotServer per the original design.

jfallows:
Also, in `whenAbortedOrFinished` you'll need to set the `activeRobotRef` back to `null`.

```
private ChannelFutureListener whenAbortedOrFinished(final ChannelHandlerContext ctx) {
        final AtomicBoolean oneTimeOnly = new AtomicBoolean();
        return new ChannelFutureListener() {
            @Override
            public void operationComplete(ChannelFuture future) throws Exception {
                if (oneTimeOnly.compareAndSet(false, true)) {
                    sendFinishedMessage(ctx);
                    activeRobotRef.compareAndSet(robot, null);
                }
            }
        };
    }
```

StCostea:
I'm sorry John, but you and David are suggesting two opposite directions. What you are now suggesting is something like I implemented before. The current version was implemented at David's suggestion. I cannot go in circle, please talk to David and take a decision on what type of implementation you want me to do.
Just to make it clear, these are the two options:
- what you just suggested - get the current running robot, add all incoming tests to its dispose future. When test finishes, only one test from the ones waiting on dispose will start and the rest will hook again on the dispose of the new running test
- what David suggested and which is the current version - don't keep the current running robot, but keep the last robot entering the queue of waiting tests. When a new test is coming, just add it to the dispose future of this last robot and update the reference to the last robot.
Please let me know which one you prefer.

dpwspoon:
I'm happy with Johns suggestion.

jfallows:
@StCostea as indicated in the comments, David and I are in agreement on the proposed approach, so kindly let us know your feedback.

Note: we should also do some minor cleanup of the `ControlServerHandler` code as part of this PR, but only after you have a functional solution in place to handle the prepare-next-before-disposed race.

StCostea:
I have added the new commit, based on John's suggestions and on the previous version I had.

PR #434, opened 2017-05-12
Wait for ABORT before close and send interrupt only once

StCostea:
Two small fixes: 
-wait for the ABORT to be sent before closing the connection
-make sure the main thread is interrupted only once

### DIFF
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
@@ -15,6 +15,7 @@
  */
 package org.kaazing.k3po.junit.rules;
 
+import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -131,21 +132,16 @@ class Latch {
     }
 
     void notifyException(Exception exception) {
-        this.exception = exception;
-        interruptTestThread();
+        this.exception = Objects.requireNonNull(exception);
+
+        if (! Thread.currentThread().equals(testThread) 
+                && testThreadInterrupted.compareAndSet(false, true)) {
+            testThread.interrupt();
+        }
     }
 
     public void setInterruptOnException(Thread testThread) {
         this.testThread = testThread;
-    }
-    
-    private void interruptTestThread() {
-        if (exception == null || testThread == null)
-            return;
-
-        if (! testThread.equals(Thread.currentThread()) && testThreadInterrupted.compareAndSet(false, true)) {
-            testThread.interrupt();
-        }
     }
 
     public Exception getException() {
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
@@ -16,6 +16,7 @@
 package org.kaazing.k3po.junit.rules;
 
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 class Latch {
 
@@ -28,6 +29,8 @@ class Latch {
     private final CountDownLatch startable;
     private final CountDownLatch finished;
     private volatile Thread testThread;
+    
+    private final AtomicBoolean exceptionOccurred = new AtomicBoolean(false); 
 
     Latch() {
         state = State.INIT;
@@ -129,19 +132,19 @@ class Latch {
 
     void notifyException(Exception exception) {
         this.exception = exception;
-        // Commented because of issue https://github.com/k3po/k3po/issues/391
-//        prepared.countDown();
-//        startable.countDown();
-//        finished.countDown();
-        if (testThread != null) {
-            if (testThread != Thread.currentThread())
-                testThread.interrupt();
+        if (testThread != null 
+                && testThread != Thread.currentThread() 
+                && exceptionOccurred.compareAndSet(false, true))
+        {
+            testThread.interrupt();
         }
     }
 
     public void setInterruptOnException(Thread testThread) {
         this.testThread = testThread;
-        if (this.exception != null) {
+        if (this.exception != null 
+                && testThread != Thread.currentThread() 
+                && exceptionOccurred.compareAndSet(false, true)) {
             testThread.interrupt();
         }
     }
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
@@ -30,7 +30,7 @@ class Latch {
     private final CountDownLatch finished;
     private volatile Thread testThread;
     
-    private final AtomicBoolean exceptionOccurred = new AtomicBoolean(false); 
+    private final AtomicBoolean testThreadInterrupted = new AtomicBoolean(false); 
 
     Latch() {
         state = State.INIT;
@@ -132,19 +132,19 @@ class Latch {
 
     void notifyException(Exception exception) {
         this.exception = exception;
-        if (testThread != null 
-                && testThread != Thread.currentThread() 
-                && exceptionOccurred.compareAndSet(false, true))
-        {
-            testThread.interrupt();
-        }
+        interruptTestThread();
     }
 
     public void setInterruptOnException(Thread testThread) {
         this.testThread = testThread;
-        if (this.exception != null 
-                && testThread != Thread.currentThread() 
-                && exceptionOccurred.compareAndSet(false, true)) {
+        interruptTestThread();
+    }
+    
+    private void interruptTestThread() {
+        if (exception == null || testThread == null)
+            return;
+
+        if (! testThread.equals(Thread.currentThread()) && testThreadInterrupted.compareAndSet(false, true)) {
             testThread.interrupt();
         }
     }
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
@@ -137,7 +137,6 @@ class Latch {
 
     public void setInterruptOnException(Thread testThread) {
         this.testThread = testThread;
-        interruptTestThread();
     }
     
     private void interruptTestThread() {
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
@@ -182,39 +182,29 @@ final class ScriptRunner implements Callable<ScriptPair> {
         controller.writeCommand(abort);
     }
 
-    public void awaitBarrier(String barrierName) throws InterruptedException {
+    public void awaitBarrier(String barrierName) throws Exception {
         if (!barriers.keySet().contains(barrierName)) {
             throw new IllegalArgumentException(String.format(
                     "Barrier with %s is not present in the script and thus can't be waited upon", barrierName));
         }
-        try {
-            controller.sendAwaitBarrier(barrierName);
-        } catch (Exception e) {
-            latch.notifyException(e);
-        }
+        controller.sendAwaitBarrier(barrierName);
         final CountDownLatch notifiedLatch = barriers.get(barrierName);
         notifiedLatch.await();
     }
 
-    public void notifyBarrier(final String barrierName) throws InterruptedException {
+    public void notifyBarrier(final String barrierName) throws Exception {
         if (!barriers.keySet().contains(barrierName)) {
             throw new IllegalArgumentException(String.format(
                     "Barrier with %s is not present in the script and thus can't be notified", barrierName));
         }
         final CountDownLatch notifiedLatch = barriers.get(barrierName);
         if (notifiedLatch.getCount() > 0) {
-            try {
-                controller.notifyBarrier(barrierName);
-            } catch (Exception e) {
-                latch.notifyException(e);
-            }
+            controller.notifyBarrier(barrierName);
         }
         notifiedLatch.await();
     }
 
-
     public void dispose() throws Exception {
-        latch.awaitFinished();
         if (controller.isConnected())
             controller.disconnect();
     }
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
@@ -214,6 +214,7 @@ final class ScriptRunner implements Callable<ScriptPair> {
 
 
     public void dispose() throws Exception {
+        latch.awaitFinished();
         if (controller.isConnected())
             controller.disconnect();
     }
--- a/junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
+++ b/junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
@@ -68,39 +68,43 @@ final class SpecificationStatement extends Statement {
             try {
                 // note: JUnit timeout will trigger an exception
                 statement.evaluate();
-            } catch (AssumptionViolatedException e) {
-
+            } catch (Throwable cause) {
+                // any exception aborts the script (including timeout)
                 if (!latch.isFinished()) {
                     scriptRunner.abort();
                 }
 
-                throw e;
-            } catch (Throwable cause) {
-                // any exception aborts the script (including timeout)
+                // wait at most 5sec for the observed script (due to the abort case)
+                // should take less than a second for K3PO to complete
+                ScriptPair scripts = null;
+                Exception scriptFutureException = null;
+                try {
+                    scripts = scriptFuture.get(5, SECONDS);
+                } catch (Exception e) {
+                    scriptFutureException = e;
+                }
+                // now that the script is stopped, handle exception
+
+                if (cause instanceof AssumptionViolatedException) {
+                    throw cause;
+                }
+
                 if (latch.hasException()) {
                     // propagate exception if the latch has an exception
                 	if (cause instanceof InterruptedException) {
                 		// take the error from the latch, it is more meaningful
                 		throw latch.getException();
                 	}
-                    throw cause;
+                    throw cause.initCause(latch.getException());
                 } else {
-                    // It is possible that the script is finished even if we get an exception
-                    // in particular a timeout may occur but the script may finish before we actually
-                    // process it the timeout, send the abort and get the result back.
-
-                    // No reason to send the abort if we are already finished.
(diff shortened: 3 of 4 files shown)


junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -129,19 +132,19 @@ boolean hasException() {
```
dpwspoon: NotifyException and setInterrrupteOnException are exactly the same? if so they should be combined

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -28,6 +29,8 @@
     private final CountDownLatch startable;
     private final CountDownLatch finished;
     private volatile Thread testThread;
+    
+    private final AtomicBoolean exceptionOccurred = new AtomicBoolean(false);
```
dpwspoon: What are you trying to prevent, setting of multiple exceptions, or sending interrupt only once, or both.

** For the multiple exceptions **

I see three spots in the code setting the exception
1) sending notify barrier, (which only occurs if it can send it on the TCP connect, i.e. it is closed)
2) sending await barrier, (which only occurs if it can send it on the TCP connect, i.e. it is closed)
3) and in latch/code execution

I believe case 1 and case 2 can only happen if there is already an error or success case (sending a notify that isn't needed).  I would, therefore, suggest we change case 1 such that exceptions are ignored.  And change case 2, such that the exception merely bypasses the waiting on the latch (don't rethrow or save it)

Making this change is good and could be done regardless of adding the AtomicBoolean above.

** For multiple interrupts **

Even with the suggested changes above, it looks like the test thread might still be interrupted multiple times (because there is one place that calls `notifyException` and a different location that calls set exception, both of which then send the interrupt) ?  If this is an issue,  then please rename this variable to testThreadInterrupted or something similar.  I would also suggest refactoring the code to extract a shared method `interruptTestThread()` or something like that and have that be the sole place that messes with AtomicBoolean.  Both the methods that send the interrupt would call this new private method.

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
```diff
@@ -214,6 +214,7 @@ public void notifyBarrier(final String barrierName) throws InterruptedException
 
 
     public void dispose() throws Exception {
+        latch.awaitFinished();
```
dpwspoon: What are you trying to fix here, and what condition or sequence is causing it to be an issue?

I'm concerned that the tests might get into a state where it blocks here forever, specifically, SpecificationStatement evaluate has a lot of catching of exceptions on latch and that would trigger a finally that calls this.  It's hard to reason if whether in all these cases that the scriptFuture Thread will still be in a good state AND latch finished will eventually be triggered.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
dpwspoon: Actually, I don't get the one and only call to `setInterruptOnException()`, in [SpecificationStatement](https://github.com/k3po/k3po/blob/7ab1dd496cda4bfb5b9d2659590746ba8d237967/junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java).  It is 

```java
      latch.setInterruptOnException(Thread.currentThread());
```

Meaning the thread is interrupting itself, which I believe is a NOOP.  Thus we can refactor `setInterruptOnException` to something like
```
     public void setTestThread(Thread testThread) {
          this.testThread = testThread
      } 
```

This should completely remove the need for the Atomic.

StCostea: `setInterruptOnException` is only setting up the connection to the main thread, the one that should be interrupted. In case there is already an exception in the latch, it will be interrupted. 

The design is not based only on the current implementation of the junit module. `setInterruptOnException` is a public method, it can be called from any place, any time. As such, it should be implemented to work properly for any future change that will call this method from other places, on any other thread.

Related to the actual fix, what I found was that in a test where k3po rule is around the gateway rule, if there was an exception from the gateway statement, then `notifyException` was called twice, once for the actual exception and a second time when an exception was thrown when k3po was trying to clean up.  For me it seems to difficult to make sure that `notifyException` is called only once from the entire flow, so it is much cleaner and simpler just to make sure that the main thread is interrupted only once. Otherwise we will have again (what I have actually see in the logs) interrupted status leaked from one test to the next one.

As for multiple exceptions, maybe we should record all of them, not only the last one, as we do now.

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
StCostea: First of all, even without this change, we are still waiting for `latch.awaitFinished()` in K3poRule. So if it will get stuck, the test will get stuck anyway. It does not matter if I'm adding one more call to `latch.awaitFinished()`.

But one of the goals of the changes I did all this year is to have that communication thread stuck only in extremely rare situations.  If you look at `ScriptRunner.call()`, there is only one exception being caught inside the loop - `SocketTimeoutException`. Any other exception will end the loop and the thread will end, with a finally that will notify the finished latch. 

The change I did in this PR is for the following reason: I have found that in a case of an exception in `base.evaluate()`, first we call `scriptRunner.abort()` and then in the finally we call `scriptRunner.dispose()`. Now that we eliminated the DISPOSE command, method `dispose()` is just closing the connection. Sometimes the closing of the connection was happening before the communication thread managed to send the ABORT and wait for the FINISHED to come. It was just like not calling abort() at all.

StCostea: As for blocked threads, that is why we have normally the timeout rule around k3po, preferably also around the gateway.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: Regarding the changes you suggest in await/notify barrier, I don't agree. If the connection is lost and we ignore the exception, then the flow in the test will continue and we will get an exception in another place. The new exception might be misleading, not reflecting the actual situation encountered. Also, in awaitBarrier, if we ignore the error we might get stuck in waiting for the latch that will never be triggered.

I don't think we should ignore exceptions, but I guess we need to change the code to properly handle them - stop communication thread and throw exception. Calling latch.notifyException might not be the best way to handle the communication error. 

I think we should open another issue for this change, to explore in more depth the changes required.

dpwspoon: Public API?  The latch code is our own...?  It can probably be (should be?) moved to internal package... Does any test use that API?? 

 I still don't get the setInterruptOnException call from the same Thread.   See JavaDoc on [interrupt()](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#interrupt()).  The only side effect of this call, when called from the current thread, is to set the interrupt flag (which could be transparently cleared depending on call stack / rules).  I don't see anywhere in the junit Statement javadoc or doc recommending this practice, and I can see how this could break other rules that might have a latch or block waiting for clean up (like what we would like to do with the Gateway Rule).

Happy to discuss the notify/await barrier, but my position is that as recommended in Code Complete, don't throw duplicate exceptions for a single cause.  Where ever the code went wrong that would cause a future await/notify barrier call to fail, (i.e. connection failure) we should be catching the exception there and displaying it properly.  If there is a case where await/notify barrier exceptions could be thrown and different then an existing exception, then it wouldn't be violating this best practice and I would change my opinion. I'd prefer doing the proper thing here in this PR, and then squashing newly visible bugs as they arise.

dpwspoon: > Related to the actual fix, what I found was that in a test where k3po rule is around the gateway rule, if there was an exception from the gateway statement, then notifyException was called twice, once for the actual exception and a second time when an exception was thrown when k3po was trying to clean up.

Awesome find, I think we can fix the issue by simplifying the code, rather than adding to it for work arounds.

junit/src/main/java/org/kaazing/k3po/junit/rules/ScriptRunner.java
dpwspoon: We need to fix all the cases where latch.awaitFinished() cause test blocking.  John and Jitu hit a case recently in debugging, and have or will be filing a bug.

We can chat in person, but there are use cases to order the rules in various orders, so I would suggest not trying to solve for what we consider common, but try to solve for all.

Your reasoning on the bug and finding it makes complete sense.  Awesome job! but I think we can implement the fix better as it adds complexity and fatal error condition where we block forever.  I would suggest trying to fix it where you observed it,  I think you are talking about this code block:

```java
            } catch (AssumptionViolatedException e) {

                if (!latch.isFinished()) {
                    scriptRunner.abort();
                }

                throw e;
            } 
...
        } finally {
            try {
                scriptRunner.dispose();
```

The catch code should probably wait for an amount of time for the finish (see the catch directly below), i.e.

```
                    try {
                        // wait at most 5sec for the observed script (due to the abort case)
                        // should take less than a second for K3PO to complete
                        ScriptPair scripts = scriptFuture.get(5, SECONDS);

                        try {
                            assertEquals("Specified behavior did not match", scripts.getExpectedScript(),
                                    scripts.getObservedScript());
                            // Throw the original exception if we are equal
                            throw cause;
                        } catch (ComparisonFailure f) {
                            // throw an exception that highlights the difference in behavior, but caused by the timeout
                            // (or
                            // original exception)
                            f.initCause(cause);
                            throw f;
                        }
```

Making the two different catch the exact same and allowing us to not special case them (simplification!!).  I haven't tried this and this is shoot from the hip suggestion, you might have to rework it.

StCostea: There are many latches and we can block on any of them (prepare, start, finished, barrier latches). Especially while debugging we can encounter some problems. If we want to avoid the k3po blocking we will have to rework the entire junit side with timeouts. From my point of view, this could be an entire refactoring work by itself.

As for the different exception handling, there is a difference between AssumptionViolatedException (that is normal and expected in a test that is failing and that should be passed further) and all the other exceptions that should be handled in k3po (or at least doing a cleanup). I don't think we can have a unified handling of the exceptions. Even in your second code block, the wait is on the result of the scriptFuture, not on a latch.
You can open an improvement issue if you really want to investigate further on this direction.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: I will clean up `setInterruptOnException` to set only the handler.

I also realized that `awaitBarrier` and `notifyBarrier` are only called from the main test thread. Because of this, the exception should not be handled using the Latch (`Latch.notifyException()` is used to communicate an exception that occurred in the communication thread to the main thread, not the other way around), but it should just be thrown. It will be caught in `evaluate()` and handled there properly.

As for the catch blocks in the statement, I'm still working on them, to try to unify the code. First I need to make sure we cover all situations.

I hope to commit a new version by EOD.

StCostea: One more thing that I see, there is no sync on the controller that sends commands to the k3po driver. In case of `notifyBarrier` and `awaitBarrier`, these are sent from the main thread, but the rest of the communication is done on the second thread. I think this might cause problems. Maybe we need a synchronize or something.

StCostea: I have added a new version with merged catch blocks. 

I also removed the catch from `awaitBarrier` and `notifyBarrier`. This might require changes in tests that are not declared to throw Exception and are using barriers. In gateway I found only 3 cases.

dpwspoon:
This looks good to me.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -129,19 +132,18 @@ boolean hasException() {
...
+        interruptTestThread();
     }
 
     public void setInterruptOnException(Thread testThread) {
         this.testThread = testThread;
-        if (this.exception != null) {
+    }
+    
+    private void interruptTestThread() {
+        if (exception == null || testThread == null)
```
dpwspoon: nit, please always include brackets with if statements

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
```diff
@@ -68,39 +68,43 @@ public void evaluate() throws Throwable {
...
                 if (!latch.isFinished()) {
                     scriptRunner.abort();
                 }
 
-                throw e;
-            } catch (Throwable cause) {
-                // any exception aborts the script (including timeout)
+                // wait at most 5sec for the observed script (due to the abort case)
+                // should take less than a second for K3PO to complete
+                ScriptPair scripts = null;
```
dpwspoon: I think this logic is correct, but the code complexity is high (mostly due to legacy problems).  If possible I would like to try to simplify it.

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
```diff
@@ -68,39 +68,43 @@ public void evaluate() throws Throwable {
             try {
                 // note: JUnit timeout will trigger an exception
                 statement.evaluate();
-            } catch (AssumptionViolatedException e) {
-
+            } catch (Throwable cause) {
```
jfallows: Can we switch this to catch `Exception` instead?
Catching `Throwable` is generally frowned upon because it catches internal JVM `Error` as well as `Exception`.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -129,19 +132,18 @@ boolean hasException() {
...
         this.exception = exception;
-        // Commented because of issue https://github.com/k3po/k3po/issues/391
-//        prepared.countDown();
-//        startable.countDown();
-//        finished.countDown();
-        if (testThread != null) {
-            if (testThread != Thread.currentThread())
-                testThread.interrupt();
-        }
+        interruptTestThread();
```
jfallows: I'd like to see the code further simplified if possible please.

If calling `Thread.interrupt()` on an already interrupted thread is a no-op, then I'd suggest the following instead...
```
if (exception != null && !currentThread().equals(testThread)) {
    testThread.interrupt();
}
```
...and remove the `testThreadInterrupted` field and the separate `interruptTestThread()` method.

Note: this is common practice to move the possibly-null variable to the argument of `equals` so that a `null` value returns `false` for `equals` without requiring the extra `!= null` check.

For example:
`"nonNullString".equals(nullableString)`
instead of
`nullableString != null && nullableString.equals("nonNullString")`

Note: also common practice to put the cheap-check first in an `if` condition:
`if (exception != null && !currentThread().equals(testThread)) ...`
instead of
`if (!currentThread().equals(testThread) && exception != null) ...`

jfallows: If it is illegal to call `notifyException` with a `null` exception, then the following would be even simpler.
```
void notifyException(Exception exception) {
    this.exception = Objects.requireNonNull(exception, "exception");

    if (!currentThread().equals(testThread)) {
        testThread.interrupt();
    }
}
```

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: I'm trying to get used to that. :) Fixed.

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
StCostea: It was like this from previous code. I changed it.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
StCostea: > If calling `Thread.interrupt()` on an already interrupted thread is a no-op

This is not always true. If `Thread.interrupt()` is called and the thread clears its interrupt status (by throwing an `InterruptedException` or by calling `Thread.interrupted()` ), next call will set again the interrupt status flag. This will make the next `await` call on any latch to throw an `InterruptedException` with no apparent reason, even in another test.
This is why I added the `testThreadInterrupted` field in the first place.

From my point of view, I think we should remove completely the whole use of interrupts. They can pass from one test to another.

junit/src/main/java/org/kaazing/k3po/junit/rules/SpecificationStatement.java
StCostea: I tried but I am not able to reduce the complexity. The only thing I could do is to remove the `else` branches to make the code more linear.

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
```diff
@@ -131,22 +132,17 @@ boolean hasException() {
     }
 
     void notifyException(Exception exception) {
-        this.exception = exception;
-        interruptTestThread();
+        this.exception = Objects.requireNonNull(exception);
+
+        if (! Thread.currentThread().equals(testThread) 
+                && testThreadInterrupted.compareAndSet(false, true)) {
```
jfallows: What breaks if the `testThreadInterrupted` part of the if condition is removed?

junit/src/main/java/org/kaazing/k3po/junit/rules/Latch.java
jfallows: Sorry, just saw this comment, which directly speaks to my most recent question.

Wouldn't it make sense to use the non-static thread method `testThread.isInterrupted()` method to determine if the test thread is interrupted (from another thread) without clearing the `interrupted` status?

StCostea: I don't understand how checking the current status will help. Can you please explain ?

This is the way it is working if we replace in the current code the `testThreadInterrupted` field with a call to `testThread.isInterrupted()` : the first interrupt will wake the main thread from the `awaitPrepared` or `awaitFinished` and will also clear the interrupted status. Thus, when the second one will come, `testThread.isInterrupted()` will show false, so it will set the status again. But this time, the main thread will not be in an await anymore, so the status will not be cleared.
It would be useful if we can check if the main thread is locked in an await. But I don't know if we can do that.

StCostea: Maybe we can use a check on Thread.getState() to check if it is in an await state. What do you think ?

PR #451, opened 2017-10-04
Fixes to allow integer and long properties to be read and written using correct byte order

cmebarrow:
1. Fixed the handling of write expression and read expression to respect the expression type (determined at runtime) rather than imposing byte array. Added tests for reading and writing numerical values, as literals and as properties using expressions.
2. Removed the TypeConverterImpl methods to coerce types to and from byte array because they should no longer be necessary.
3. Made various other fixes required by 2.

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/visitor/GenerateConfigurationVisitor.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/visitor/GenerateConfigurationVisitor.java
@@ -498,7 +498,7 @@ public class GenerateConfigurationVisitor implements AstNode.Visitor<Configurati
 
         @Override
         public MessageEncoder visit(AstExpressionValue<?> value, Void parameter) {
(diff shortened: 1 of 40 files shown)


driver/src/main/java/org/kaazing/k3po/driver/internal/behavior/visitor/GenerateConfigurationVisitor.java
```diff
@@ -498,7 +498,7 @@ public Configuration visit(AstWriteValueNode node, State state) {
 
         @Override
         public MessageEncoder visit(AstExpressionValue<?> value, Void parameter) {
-            Supplier<byte[]> supplier = () -> value.getValue(byte[].class);
+            Supplier<Object> supplier = () -> value.getValue(Object.class);
             return new WriteExpressionEncoder(supplier, value.getExpression());
```
jfallows: Suggest `return new WriteExpressionEncoder(value::getValue, value.getExpression);` instead.

PR #456, opened 2017-10-23
TLS ALPN negotiation

jitsni:
Setting SSLParameters#setApplicationProtocols() to start ALPN negotiation.
Since the call is only available in Java 9, it is being invoked using reflection.

### DIFF
--- a/driver/src/main/java/org/kaazing/k3po/driver/internal/ext/tls/bootstrap/TlsClientChannelSink.java
+++ b/driver/src/main/java/org/kaazing/k3po/driver/internal/ext/tls/bootstrap/TlsClientChannelSink.java
@@ -32,6 +32,8 @@ import static org.kaazing.k3po.driver.internal.netty.channel.Channels.fireInputS
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.security.KeyStore;
 import java.security.SecureRandom;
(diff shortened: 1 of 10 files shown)


driver/src/main/java/org/kaazing/k3po/driver/internal/ext/tls/bootstrap/TlsClientChannelSink.java
```diff
@@ -129,6 +128,7 @@ protected void connectRequested(ChannelPipeline pipeline, ChannelStateEvent evt)
         File trustStoreFile = tlsConnectConfig.getTrustStoreFile();
         char[] keyStorePassword = tlsConnectConfig.getKeyStorePassword();
         char[] trustStorePassword = tlsConnectConfig.getTrustStorePassword();
+        String[] applicationProtocol = tlsConnectConfig.getApplicationProtocols();
```
jfallows: `applicationProtocol` -> `applicationProtocols`

driver/src/main/java/org/kaazing/k3po/driver/internal/ext/tls/bootstrap/TlsServerChannelSink.java
```diff
@@ -97,6 +97,7 @@ protected void bindRequested(ChannelPipeline pipeline, ChannelStateEvent evt) th
             File trustStoreFile = tlsConnectConfig.getTrustStoreFile();
             char[] keyStorePassword = tlsConnectConfig.getKeyStorePassword();
             char[] trustStorePassword = tlsConnectConfig.getTrustStorePassword();
+            String[] applicationProtocol = tlsConnectConfig.getApplicationProtocols();
```
jfallows: `applicationProtocol` -> `applicationProtocols`

driver/src/test/java/org/kaazing/k3po/driver/internal/tls/TlsIT.java
```diff
@@ -44,6 +45,16 @@ public void shouldEstablishConnection() throws Exception {
         k3po.finish();
     }
 
+    @Ignore("Needs JDK 9 to run as it uses SSLParameters#setApplicationProtocols()")
```
jfallows: Can we unignore this by building `k3po` with `JDK 9` but targeting `1.8` for `source` and `target` so it runs on `Java 8` too?

